PR #14, opened 2023-06-20
chore: suppress r5py warning prints in CLI

### LINKED ISSUE #13: pytest r5py java warnings
**Describe the bug**

r5py java warnings are printed in the CLI after running pytest (and passing). This is quite inconvenient since it clogs up the CLI and hides the result of the pytest run.

**Environment**

 - OS: macOS
 - Python version: 3.9.13
 - Java version: 11

**Steps to reproduce**

r5py warnings would be suppressed/not clogging up the CLI

**Screenshots**

None

**Test data/code**

N/A

**Additional context**

None

ethan-moss:
## Description

Modified pytest config in pyproject.toml to not run `faulthandler` by default.

`faulthandler` is used to traceback segmentation/timeout issues. Running this 'by default' was introduced in pytest at v5.0.0. The "-p no:faulthandler" switch runs pytest as normal without `faulthandler` being active. This is controlled in the `pyproject.toml` config, so it becomes the default behaviour. See [pytest docs](https://docs.pytest.org/en/7.1.x/how-to/failures.html#fault-handler) for more info. Note: this strategy is also adopted by `r5py`.

Fixes #13 

## Motivation and Context

r5py warnings during pytest run clutter CLI which hides the results of the pytest run (message is too long). This correction suppresses those warning messages.

## Type of change

## How Has This Been Tested?

`pytest --runsetup` runs as expected, with r5py warnings cluttering CLI. Performing dummy tweaks to the assertion statements causes the test to fail, as expected (these dummy tweaks aren't committed).

Test configuration details:
* OS: macOS
* Python version: 3.9.12
* Java version: 11
* Python management system: conda

## Checklist:

## Additional comments

None

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,7 @@ profile = "black"
 # `pytest` configurations
 [tool.pytest.ini_options]
 addopts = [
+    "-p no:faulthandler",
     "-vv",
     "--doctest-modules"
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,6 @@ profile = "black"
 # `pytest` configurations
 [tool.pytest.ini_options]
 addopts = [
-    "-p no:faulthandler",
     "-vv",
     "--doctest-modules"
 ]
--- a/tests/test_setup.py
+++ b/tests/test_setup.py
@@ -8,7 +8,6 @@ TODO: make this run 'on request' only.
 
 import pytest
 import os
-from r5py import TransportNetwork
 
 
 class TestSetup:
@@ -35,6 +34,10 @@ class TestSetup:
     @pytest.mark.setup
     def test_r5py_setup(self) -> None:
         """Check development environment will cope with r5py requirements."""
+        # this import is only here to prevent java messages being printed to
+        # the CLI when set-up pytests are not run
+        from r5py import TransportNetwork
+
         # search the ext dir for pbf & gtfs
         search_pth = os.path.join("tests", "data")
         foundf = os.listdir(search_pth)


r-leyshon:
I can see this is fresh as CI still running. Haven't got my head around the fault handler thing. The only thing I would caution here is that we want to blanket suppress the java warnings for sure, but we still need to ensure pythonic warnings are captured in the output of pytest. Hopefully this approach has considered this - happy to talk through tomorrow!

ethan-moss:
Yeah, I've done some quick tests which convinced me that the standard python error raising still works as expected (they dummy tests weren't committed, just temp checks):
- Placed dummy `raise ValueError`, `raise FileNotFoundError`, `raise DeprecationWarning`, and `raise Exception` in `test_r5py_setup()`:
    - All statements still caused a pytest failure (with the exceptions raised with an accompanying error message)
- Using `logging` and writing DEBUG, INFO, WARNING and ERROR messages to the root logger:
    - All logger statements are suppressed, but this is standard behaviour even without the `-p no:faulthandler` switch. The `-rP` switch is needed to display prints/logs when the test passes.

This strategy is also adopted in r5py's pytest config. That's not to say we shouldn't be convinced by ourself that this switch's behaviour is OK before we merge though.

ethan-moss:
Reading this [python blog post](https://blog.python.org/2011/05/new-faulthandler-module-in-python-33.html), it suggests `faulthander` is used for debugging. This makes me think it's not a error catcher or handler as such, but more a helpful util that allows packages to give some useful debugging feedback if segmentation or 'hanging' issues arise. In this case, those sorts of issues should be pretty clear (pytest would hang, for example) - not ideal since we wouldn't have helpful debugging pointers, but it wouldn't be hidden issue. If further debugging is needed, it can be turned on using the CLI switch as needed. What do you think?  Maybe there is a way to have this switch for a single test only... we could take a look.

r-leyshon (commented):
@ethan-moss has confirmed that pythonic errors propagate with no:faulthandler. I have confirmed warnings.warn() propagate as needed.

I've not fully understood a few of the comments on the PR, so suggest discussing as a team prior to merging. But so far I would say this meets our needs - appears to silence Java noise while allowing the python warnings & errors through at pytest.

r-leyshon:
> Reading this [python blog post](https://blog.python.org/2011/05/new-faulthandler-module-in-python-33.html), it suggests `faulthander` is used for debugging. This makes me think it's not a error catcher or handler as such, but more a helpful util that allows packages to give some useful debugging feedback if segmentation or 'hanging' issues arise. In this case, those sorts of issues should be pretty clear (pytest would hang, for example) - not ideal since we wouldn't have helpful debugging pointers, but it wouldn't be hidden issue. If further debugging is needed, it can be turned on using the CLI switch as needed. What do you think? Maybe there is a way to have this switch for a single test only... we could take a look.

Let's discuss this with @SergioRec

r-leyshon:
alternatively, we move the java integration tests to their own workflow file and tolerate the java warnings and less performant test (graph building notably slow)

ethan-moss:
@r-leyshon i've changed the strategy to one where we are now keeping `faulthandler` active and instead the offending import statement is moved to be within the test itself (as a one off). Now, without the `--runsetup` flag the Java messages no longer arise (as expected). These messages still appear when the flag is used, but this is tolerable given we wouldn't run with this flag often.

r-leyshon (approved):

PR #33, opened 2023-07-24
chore: CI run on all branches

### LINKED ISSUE #31: blanket CI
Discovering dependency-related issues on PR. Better to have CI setup on every branch in order to find these problems early. No need to limit CI runs on open repo as no cost incurred.

DoD:
- [ ] CI runs on pushes to all branches and PR to dev.

r-leyshon:
Really quick CI patch to:

* have CI run on all branch pushes by default
* CI runs on PR to dev and main

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -5,7 +5,6 @@ name: Python package
 
 on:
   push:
-    branches: [ "dev", "main" ]
   pull_request:
     branches: [ "dev", "main" ]
 


ethan-moss (approved):
This change looks good to me according to the [GitHub docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions) - the removal of the `branches` filter should result in triggering runners on any push to the remote. This should help us catch GitHub action bugs much sooner. Thanks!

PR #100, opened 2023-08-23
Adjust check_parent_dir_exists() to fix bugs

### LINKED ISSUE #60: `01-validation-gtfs.py` does not use `create_out_parent` arg
**Describe the bug**

When running `pipeline/gtfs/01-validation-gtfs.py`, the following error is raised:
```
  File /transport-network-performance/src/transport_performance/pipeline/gtfs/01-validate-gtfs.py, line 87, in <module>
    feed.viz_stops(out_pth=POINT_MAP_PTH)
  File /transport-network-performance/src/transport_performance/gtfs/validation.py", line 202, in viz_stops
    _check_parent_dir_exists(
  File "/transport-network-performance/src/transport_performance/utils/defence.py", line 39, in _check_parent_dir_exists
    raise FileNotFoundError(
FileNotFoundError: Parent directory outputs/gtfs/validation not found on disk.
```

This is because the output directory did not exist at the time and `viz_stops()` does not use it's `create_out_parent` arg in this pipeline script. Adding this in, and potentially adding in a corresponding flag in the config .toml file, will fix the issue as below (needs implementing on lines [87](https://github.com/datasciencecampus/transport-network-performance/blob/dev/pipeline/gtfs/01-validate-gtfs.py#L87) and [94](https://github.com/datasciencecampus/transport-network-performance/blob/dev/pipeline/gtfs/01-validate-gtfs.py#L94))

```python
feed.viz_stops(out_pth=POINT_MAP_PTH, create_out_parent=True)
```

**Environment**

 - OS: macOS
 - Python version: 3.9.13
 - Java version: 11

**Steps to reproduce**

Run the pipeline script and save the outputs to file.

**Screenshots**

None, error message shared above.

**Test data/code**

None

**Additional context**

None

### LINKED ISSUE #61: `_check_parent_dir_exists()` does not create a directory when `create=True`
**Describe the bug**

Running [`_check_parent_dir_exists()`](https://github.com/datasciencecampus/transport-network-performance/blob/96f48a9588b8e119812fad18e21cb8481b325bc5/src/transport_performance/utils/defence.py#L31) does not create a directory when `create=True` AND the requested path is a new parent folder with a child subfolder. This is because `os.mkdir()` does not create new folders recursively. Here is an example: 

```bash
>>> import os
>>> os.mkdir("outputs/gtfs") # works OK, so delete folder and try with grandchild
>>> os.mkdir("outputs/gtfs/validation")
FileNotFoundError: [Errno 2] No such file or directory: 'outputs/gtfs/validation'
```
[StackOverflow](https://stackoverflow.com/questions/6004073/how-can-i-create-directories-recursively) suggests using [`os.makedirs()`](https://docs.python.org/3/library/os.html#os.makedirs) :
```bash
>>> os.makedirs("outputs/gtfs/validation") # grandchild created as expected.
```
This could be dropped in directly to replace `os.mkdir` [link to code line](https://github.com/datasciencecampus/transport-network-performance/blob/15-gtfs-validation-pipeline/src/transport_performance/utils/defence.py#L36): 
```python
def _check_parent_dir_exists(pth, param_nm, create=False):
	_is_path_like(pth, param_nm)
	parent = os.path.dirname(pth)
	if not os.path.exists(parent):
		if create:
			os.makedirs(parent)
			print(f"Creating parent directory: {parent}")
	else:
		raise FileNotFoundError(
			f"Parent directory {parent} not found on disk."
		)

	return None
```

Or the `exist_ok` argument could be used to handle the whole 'does the folder already exist' if/else statement internally.

This aspect should also be covered by a unit test (in `test_defense.py`) to prevent reocurrence.

**Environment**

 - OS: macOS
 - Python version: 3.9.13
 - Java version: 11

**Steps to reproduce**

New folder is generated regardless of how many new sub-directories need to be created.

cbrowning04:
## Description

Fixes #60 
Fixes #61

## Motivation and Context

The motivation behind this PR is to address bugs introduced in the `_check_parent_dir_exists()` function
in src/transport_preformance/utils/defences.py. 

## Type of change

## How Has This Been Tested?

I have used unit tests to ensure that directories are being successfully created. This includes creating single parent directories and also multiple parent directories.  
Tests have also been completed to ensure that when `create=False` in `_check_parent_dir_exists()`, an error is thrown if the parent directory does not already exist.  
Further tests have been created to ensure that file paths including any number of '\' characters are processed correctly through converting to forward slash and then converting into an os specific path object.  
These tests were completed on both Windows and macOS.

Test configuration details:
Test 1:  
* OS: Windows 10
* Python version: 3.9.13
* Java version: N/A

Test 2:  
* OS: macOS Big Sur 11.6
* Python version: 3.9.13
* Java version: N/A
* Python management system: virtualenv

## Advice for reviewer

'macOS' tests were completed on my personal machine (off network, non work related).  
Also, github actions passed the tests on both Ubuntu and MacOS.
## Checklist:

### DIFF
--- a/src/transport_performance/utils/defence.py
+++ b/src/transport_performance/utils/defence.py
@@ -35,7 +35,7 @@ def _check_parent_dir_exists(pth, param_nm, create=False):
     parent = os.path.dirname(pth)
     if not os.path.exists(parent):
         if create:
-            os.mkdir(parent)
+            os.makedirs(parent)
             print(f"Creating parent directory: {parent}")
         else:
             raise FileNotFoundError(
--- a/src/transport_performance/utils/defence.py
+++ b/src/transport_performance/utils/defence.py
@@ -32,7 +32,6 @@ def _check_parent_dir_exists(pth, param_nm, create=False):
     _is_path_like(pth, param_nm)
     # replace back slashes to ensure consistency
     pth = pth.replace("\\", "/").replace(repr("\\"), "").replace("'", "")
-    pth = pth.replace("\\", "/")
     # convert path to the correct OS specific format
     pth = pathlib.Path(pth)
     # realpath helps to catch cases where relative paths are passed in main
--- a/src/transport_performance/utils/defence.py
+++ b/src/transport_performance/utils/defence.py
@@ -30,6 +30,8 @@ def _is_path_like(pth, param_nm):
 
 def _check_parent_dir_exists(pth, param_nm, create=False):
     _is_path_like(pth, param_nm)
+    # convert path to the correct OS specific format
+    pth = pathlib.Path(pth)
     # realpath helps to catch cases where relative paths are passed in main
     pth = os.path.realpath(pth)
     parent = os.path.dirname(pth)
--- a/src/transport_performance/utils/defence.py
+++ b/src/transport_performance/utils/defence.py
@@ -2,10 +2,14 @@
 import pathlib
 import numpy as np
 import os
+from typing import Union
 
 
 def _is_path_like(pth, param_nm):
     """Handle path-like parameter values.
+
+    It is important to note that paths including backslashes are not accepted,
+    with forward slashes being the only option.
 
     Parameters
     ----------
@@ -27,11 +31,49 @@ def _is_path_like(pth, param_nm):
     if not isinstance(pth, (str, pathlib.Path)):
         raise TypeError(f"`{param_nm}` expected path-like, found {type(pth)}.")
 
+    if not isinstance(pth, pathlib.Path):
+        if "\\" in repr(pth):
+            raise ValueError(
+                "Please specify string paths with single forward"
+                " slashes only."
+                f" Got {repr(pth)}"
+            )
 
-def _check_parent_dir_exists(pth, param_nm, create=False):
+
+def _check_parent_dir_exists(
+    pth: Union[str, pathlib.Path], param_nm: str, create: bool = False
+) -> None:
+    """Check if a files parent directory exists.
+
+    The parent directory in this case will be the second layer. If only a
+    single layer is passed, the parent directory will be assumed to exist as
+    it will be the current working directory.
+
+    Parameters
+    ----------
+    pth : Union[str, pathlib.Path]
+        The path to the file who's parent dir is being confirmed to exist. The
+        function will consider the second layer of the path to be the parent
+        directory.
+    param_nm : str
+        The name of the parameter for the path. This is used in the error
+        message within is_path_like()
+    create : bool, optional
+        Whether or not to create the parent directory if it does not already
+        exist, by default False
+
+    Returns
+    -------
+    None
+
+    Raises
+    ------
+    FileNotFoundError
+        An error is raised if the parent directory could not be found and also
+        the create parameter is False.
+
+    """
     _is_path_like(pth, param_nm)
-    # replace back slashes to ensure consistency
-    pth = str(pth).replace("\\", "/").replace(repr("\\"), "").replace("'", "")
     # convert path to the correct OS specific format
     pth = pathlib.Path(pth)
     # realpath helps to catch cases where relative paths are passed in main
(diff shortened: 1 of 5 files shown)


src/transport_performance/utils/defence.py
```diff
@@ -30,12 +30,16 @@ def _is_path_like(pth, param_nm):
 
 def _check_parent_dir_exists(pth, param_nm, create=False):
     _is_path_like(pth, param_nm)
+    # replace back slashes to ensure consistency
+    pth = pth.replace("\\", "/").replace(repr("\\"), "").replace("'", "")
```
cbrowning04: This isn't ideal but it's the only way to ensure paths are consistent and also that paths including singular \ (escape char) are correctly processed. It also provides the leniency for the users path to include _n_ number of '\\' in a path.  
An example of this includes `out_pth="D:\\\dir1\dir2\\dir3\dir4\file.html"`

src/transport_performance/utils/defence.py
```diff
@@ -31,7 +31,7 @@ def _is_path_like(pth, param_nm):
 def _check_parent_dir_exists(pth, param_nm, create=False):
     _is_path_like(pth, param_nm)
     # replace back slashes to ensure consistency
-    pth = pth.replace("\\", "/").replace(repr("\\"), "").replace("'", "")
+    pth = str(pth).replace("\\", "/").replace(repr("\\"), "").replace("'", "")
```
cbrowning04: This isn't ideal but it's the only way to ensure paths are consistent and also that paths including singular \ (escape char) are correctly processed. It also provides the leniency for the users path to include n number of '\' in a path.
An example of this includes out_pth="D:\\\dir1\dir2\\dir3\dir4\file.html"

src/transport_performance/utils/defence.py
```diff
@@ -30,12 +30,16 @@ def _is_path_like(pth, param_nm):
 
 def _check_parent_dir_exists(pth, param_nm, create=False):
```
SergioRec: Added to technical debt #103 : this function needs documentation. Behaviour is not as I'd expect. It will get the parent directory of whatever path is fed into the function, so if called like this, it will not throw any error even though the directory does not exist:

```
path1 = os.path.join(here(), 'no_folder')
_check_parent_dir_exists(path1, 'n', create=False)
```

It will however throw an error if something else was appended to the path:
```
path2 = os.path.join(here(), 'no_folder//no_file.txt')
_check_parent_dir_exists(path2, 'n', create=False)
```

src/transport_performance/utils/defence.py
SergioRec: When I pass your example path into the function, I get a wrong output. This is what's parsed using the `replace` line:

```
>>>path = "D:\\dir1\dir2\dir3\dir4\file.html"
>>>str(path).replace("\\", "/").replace(repr("\\"), "").replace("'", "")

'D:/dir1/dir2/dir3/dir4\x0cile.html'
```

So it seems it is escaping the "f" character ("\f" is a [FF character](https://stackoverflow.com/questions/44955635/how-to-print-ff-form-feed-character) which is printed as "\x0c"). This may be an issue in macOS only so you might not be able to reproduce in Windows.

SergioRec: Discussed with @ethan-moss and suggested that we just do some logic for this check:
- If it's a Path object, use it as is.
- If it's a string, check that it only uses single forward slashes, if not throw an error requesting the correct format.

This could be added into the `_is_path_like` function.

tests/utils/test_defence.py
```diff
@@ -60,3 +63,62 @@ def test_check_parent_dir_exists_defence(self):
...
+        # test creating the parent directory (2 levels)
+        _check_parent_dir_exists(
+            pth=os.path.join(
+                tmp_path, "test_dir_multi", "test_dir_multi", "test.html"
+            ),
+            param_nm="test_prm",
+            create=True,
+        )
+
+        assert os.path.exists(
```
SergioRec, line 131: Tested this assertion by changing `makedirs` to `mkdir` so no subfolders were created. The test fails, but I get a `FileNotFoundError: [Errno 2] No such file or directory` instead of the `AssertionError`. I think this is because if no directories are created, the function will fail before getting to the assertion. So in this case perhaps testing for this specific error with a `with FileNotFoundError` context?

SergioRec (changes requested):
`makedirs` works as expected. Small issue in test for creation of subdirectories. Proposed alternative solution for backslashes issue.

SergioRec (approved):
#60 and #61 are fixed now.

Issue #105 added as technical debt.

PR #101, opened 2023-08-23
CRS unit recognition and better convex hull calculations

### LINKED ISSUE #64: `viz_stop()` convex hull area incorrect when "km" units are used
**Describe the bug**

Area calculation in `_create_map_title_text()` is incorrect when "km" units are set - m2/Km2 units are inverted by the if statement:
    - Getting 175,465,852,248km2 for a near rectangular convex cull covering England and Wales when instantiating `GtfsInstance` with `units="km"`- should be ~250,000Km2 (based on Wales/England width and height being approx 500Km by 500Km when measuring the distances). 

Fix:
Area units after calculating area will always be in m2 if the CRS units are in m:
    - this is true even if the GTFS units are in Km - the CRS units define the area's unit (CRS EPSG:27700 is in m).
    - Therefore, I think the `if` statement isn't needed and the following code snippet will always return the area in Km if the CRS unit is in m.
    - Fix could look like this [replacing lines](https://github.com/datasciencecampus/transport-network-performance/blob/96f48a9588b8e119812fad18e21cb8481b325bc5/src/transport_performance/gtfs/validation.py#L73-L76)(tested locally and it works):
```
hull_km2 = gdf.to_crs(geom_crs).area / 1000000
```
    - But would also need to make sure the unit of the CRS is also m.

**Environment**

 - OS: macOS
 - Python version: 3.9.13
 - Java version: 11

**Steps to reproduce**

Area remains consistent regardless of the GTFS units.

cbrowning04:
## Description

Fixes #64

## Motivation and Context

The reasoning for this PR is to ensure that convex hull areas are being correctly calculated. More specifically, when units are 'km' in a `GtfsInstance` object. These changes lead to the convex hull area being calculated correctly depending on the CRS units, rather than the `units` attribute of a `GtfsInstance` object.

## Type of change

## How Has This Been Tested?

This has been tested through unit tests and also local testing. Multiple CRS have been tested and have yielded the correct results. The convex hull area also returned as the correct value when setting units to 'km' for the GTFS instance.

Test configuration details:
* OS: Windows 10
* Python version: 1.9.13
* Java version: N/A

## Checklist:

### DIFF
--- a/src/transport_performance/gtfs/validation.py
+++ b/src/transport_performance/gtfs/validation.py
@@ -8,6 +8,7 @@ import datetime
 import numpy as np
 import os
 import inspect
+import warnings
 
 from transport_performance.gtfs.routes import scrape_route_type_lookup
 from transport_performance.utils.defence import (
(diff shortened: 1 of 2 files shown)


ethan-moss (approved):
Thanks for working on this fix - it looks good to me!

I've tested this locally to show:
- The convex hull's area calculation is now (correctly) decoupled from the units of the GTFS file: I tried both "m" and "km" and get the same area in Km2 regardless of the GTFS units
- The unit conversion (m/km) is now (correctly) done on the CRS's units <--- this is a nice touch and a much better way of solving the problem, nicely done!
- A `UserWarning` was triggered when the CRS requested has units not in "m" or "km". This is a nice feature, since together with the `geopandas` warning it's clear a difference CRS needs to be provided for the area calculation:
- The added unit test passes, without new warnings, and tests this expected functionality:

Note: the `UserWarning` checked by `pytest` will be the one triggered by `geopandas` since this is the one raised first. So technically, the unit test wouldn't check the `UserWarning` raised by this function which could be a silent failure. Instead, i've changed the `UserWarning` to a `ValueError` and have updated the unit test to suit. It now checks for both the `geopandas` `UserWarning` and this `ValueError` at the same time. It also now stops at the point the `ValueError` is raised, as opposed to continuing to export the .html output when `viz_stop()` is called <--- this last point I think it important, since we know the convex hull's area estimate would be incorrect, so lets not export something in this case.

All other unit tests pass as expected too <--- diff is small, and does not impact other files, so all good. We can merge this fix to close out the bug. Thanks again!

PR #112, opened 2023-08-31
HTML outputs for GTFS data.

### LINKED ISSUE #45: GtfsInstance Summary Plotting Functions
### Description
Functions to plot summarised route/trip data from the GTFS file. This data is obtained by calling summarise_trips() or summarise_routes() on a GtfsInstance object. Plots should be created using plotly.

### Acceptance Criteria
- [x] Working code written
- [x] Unit test written
- [x] Flexible for trip and route summaries
- [x] Interactive plots allowing for route types to be selected
- [x] can be outputted as a html object
- [x] raise errors if `summarise_routes()` or `summarise_trips()` is yet to be called while attempting to plot.

### LINKED ISSUE #46: HTML Outputs Showing GTFS Validity Information
### Description
Convert the outputs of the GTFS pipeline (informative outputs, not visualisation or raw data) and present them on a static HTML page. The main feature of this page will be the ability to traverse the errors identified through gtfs_feed.validate()

### Acceptance Criteria
- [x] Working code written
- [x] Unit test written
- [x] HTML page displays results of feed.validate()

### LINKED ISSUE #47: HTML Outputs Showing GTFS stops/routes/convex hull
### Description
Move the outputs currently generated from `GtfsInstance.viz_stops()` onto a static HTML output

### Acceptance Criteria
- [x] Working code written
- [x] Unit test written
- [x] HTML page displays results of `GtfsInstance.viz_stops()`

### LINKED ISSUE #48: HTML Outputs Showing GTFS Summaries (Stops+Routes)
### Description
Display the outputs from `GtfsInstance.summarise_trips()` and `GtfsInstance.summarise_routes()` on a static HTML page

### Acceptance Criteria
- [x] Working code written
- [x] Unit test written
- [x] HTML page displays results of `GtfsInstance.summarise_trips()` and `GtfsInstance.summarise_routes()` 
- [x] Results are interactive (using plotly)

### LINKED ISSUE #49: Convert GTFS summaries to singular index df's and normalise column names
### Description
In order to make the summary dataframes more user-friendly and also easier to work with, they should be changed from a multi index to a singular index.  
Differing numpy versions also provide different operator column names (amin/min, amax/max). This should be normalised, along with column order.  
### Acceptance Criteria
- [x] Working code written
- [x] Unit tests updated
- [x] All summarised outputs have a singular index
- [x] Column order/names are consistent

### LINKED ISSUE #59: Update CSS in gtfs-html-outputs: Hovering over sidebar item
### Description
Text only changes colour when it is being hovered over, rather than when the surrounding box is being hovered over.

### Acceptance Criteria
- [ ] When surrounding box changes colour, the text also changes colour

cbrowning04:
## Description

closes #47 
closes #59 
closes #48 
closes #46 
closes #45 
closes #49 

## Motivation and Context

This PR intends to allow for HTML reports to be created on GTFS datasets. The key functionality added through this PR is the `gtfs::validation::GtfsInstance().html_report()` method.

## Type of change

## How Has This Been Tested?

Test configuration details:
* OS: Windows 10
* Python version: 3.9.13
* Java version: N/A
* Python management system: conda

## Checklist:


src/transport_performance/utils/defence.py
```diff
@@ -169,6 +170,50 @@ def _bool_defence(some_bool, param_nm):
...
+
+
+# main use case for this is to avoid
+# complexity limits in
+# GtfsInstance._plot_summary()
+def _string_and_nonetype_defence(some_value, param_nm):
+    if not isinstance(some_value, (str, type(None))):
+        raise TypeError(
+            f"'{param_nm}' expected type str. Found type {type(some_value)}"
+        )
```
r-leyshon: @CBROWN-ONS interested in this function. It's like `_string_defence` but additionally passes for None. I can see that `GtfsInstance._plot_summary` will play ball with x and y labels or without - nice. I can see 2 ways forward for this utility:
1. Update the TypeError message to mention nonetype.
2. Scrap this function, moving the logic into `_string_defence` with an additional parameter `raise_on_nonetype = True`.
Also worth considering whether any of the other defence utils would benefit from the same approach. Do you have a view on which would be better?

tests/gtfs/test_validation.py
```diff
@@ -580,3 +602,233 @@ def test_summarise_routes_on_pass(self, gtfs_fixture):
...
+        ), "Failed to save summary in HTML"
+        assert os.path.exists(
+            os.path.join(tmp_path, "save_test", "test_image.png")
+        ), "Failed to save summary as a PNG"
+
+        # save test for HTML with an invalid file extension
+        with pytest.warns(
+            UserWarning,
+            match=re.escape(
+                "HTML save requested but accepted type not specified.\n"
```
r-leyshon: @CBROWN-ONS by accepted type, do you mean the user asked for html as `save_html=True` but wrong file extension provided? If I've interpreted the behaviour correctly, suggest a UserWarning like:
"`save_html` is True but invalid file extension passed to `save_pth`."
Alternatively, you could check the extension matches the correct parameter `save_html` or `save_image` and stomp on the bad file extension with the required one. Good idea to raise a warning in that event, too. Would you rather adjust the warning message or behaviour though?

tests/gtfs/test_validation.py
```diff
@@ -580,3 +602,233 @@ def test_summarise_routes_on_pass(self, gtfs_fixture):
...
+                " Found l of type <class 'str'>."
+            ),
+        ):
+            current_fixture._plot_summary(
+                current_fixture.daily_route_summary,
+                "route_count_mean",
+                orientation="l",
+            )
+
+    @pytest.mark.filterwarnings("ignore::UserWarning")
```
r-leyshon: @CBROWN-ONS this one needs some thought. Ideally, we want to test the behaviour of these warnings rather than filter them. Did you try it? I have found this to be unexpectedly hacky in Python and if it's way too much work to implement, will potentially spin a separate issue up.

src/transport_performance/utils/defence.py
cbrowning04: Regarding point 1, I definitely think this should be the case.  
Moving onto point 2, I feel that we could pass a parameter to the defences to allow None types. This would also simplify the defence.py file as there wouldn't have to be multiple functions for each datatype defence.  Whether or no there are any other use cases throughout the package I'm not sure.

tests/gtfs/test_validation.py
cbrowning04: Essentially there isn't really a good answer for this as the user can save both a HTML output and an image at the same time, leading to conflicting file extension options. Because of this it essentially is going to raise a warning each time you choose to save as both.  
A possible adjustment to the error message is to recognise the cases where both `save_image` and `save_html` == `True`. Another option is to generalise the path to only accept paths without file extensions, and then possibly add another param for image file extensions (where there are multiple options).

tests/gtfs/test_validation.py
cbrowning04: The main rationale for this was that the functionality of saving both an image and a html file at one time would always raise a warning as you are guaranteed not to have specified a correct file type (e.g., .PNG) for both html and image saves. This only appears as a UserWarning to inform the user, however the overall functionality remains consistent and file extensions will always be adjusted to a valid type.  
Without filtering these warnings, they would show up in the test report, indicating that there was a warning  (or something wrong with tests) when in reality there wasn't any major issue.  
I'm not sure how to approach this but it's possible that the solution I mentioned in the above comment may help to resolve this.

src/transport_performance/utils/defence.py
r-leyshon: Actually, can this replace all of the defence utilities introduced here? (Haven't pushed this up yet but I can do if you want to test it out).
```
def _type_defence(some_object, param_nm, types) -> None:
    """Internal defence checking utility. Can handle NoneType.

    Parameters
    ----------
    some_object : Any
        Object to test with isinstance.
    param_nm : str
        A name for the parameter. Useful when this utility is used in a wrapper
        to inherit the parent's parameter name and present in error message.
    types : type or tuple
        A type or a tuple of types to test `some_object` against.
    Raises
    ------
    TypeError
        `some_object` is not of type `types`. 

    """
    if not isinstance(some_object, types):
        raise TypeError(
            f"`{param_nm}` expected {types}. Got {type(some_object)}"
            )
    
    return None
```

I would suggest this one func could replace:
* _bool_defence
* _string_defence
* _integer_defence
* _dict_defence
* _dataframe_defence
* _string_and_nonetype_defence (by using it with a tuple of types, like `_type_defence("some string", "x label", (str, type(None))`)

cbrowning04: I thought of doing this before and I agree that it's a good idea.  
It also removes the need to add any additional type-checking defences in the future. I'm happy for you to push this assuming that it has been tested locally and retains the same behaviour as the previous defences.  

You'd have to update all instances of usage before pushing this, including functions impacted by `_bool_defence()` which was already present in `defences.py`.  

Overall, I'd be happy with this change, much better than several smaller functions!

r-leyshon: I just pushed a refactored version of this branch and raised a PR for your consideration @CBROWN-ONS :
https://github.com/datasciencecampus/transport-network-performance/pull/123

This branches off gtfs-html-new and wants to merge back to that branch.

tests/gtfs/test_validation.py
r-leyshon: @CBROWN-ONS OK, how about:
* ask the user where to save the map - in which directory.
* Give them a filename, can use an f string of your choice with today's datestamp to avoid overwriting on multiple runs
* dependent on status of `save_image` / `save_html` you can ensure they have the expected file extension without needing to check for janky user behaviour. Although, what happens if both of those params are set to False?
* Consider going with an `output_format` param instead, taking string or list of strings with following acceptable values = ["html", "svg", "pdf"... so on]

Does that sound like we could avoid needing to raise warnings? They're tricky to test, right?

CBROWN-ONS: Have you seen the behaviour on lines 627 to 667 in gtfs.validation.py? It essentially ensure that when save_html=True, .html is specified, and the same for save_image (with it's respective accepted formats). This sounds similar to what you're suggesting unless I'm understanding incorrectly.

CBROWN-ONS: However the output_format suggestion sounds like a good implementation.

cbrowning04: @r-leyshon Essentially it already checks for the correct file extension, and imputes one if the users file extension is deemed invalid. In theory, the warnings could be converted to print statements as the functionality remains the same. No matter what, the figure will be saved correctly.

```python
 # save the plot if specified (with correct file type)
if save_html:
    if ext.lower() != ".html":
        warnings.warn(
            (
                "HTML save requested but accepted type "
                "not specified.\n"
                "Accepted image formats include ['.html']\n"
                "Saving as .html"
            ),
            UserWarning,
        )
    plotly_io.write_html(
        fig=fig,
        file=os.path.normpath(main + ".html"),
        full_html=False,
    )

if save_image:
    valid_img_formats = [
        ".png",
        ".pdf",
        "jpg",
        "jpeg",
        ".webp",
        ".svg",
    ]
    if ext.lower() not in valid_img_formats:
        warnings.warn(
            (
                "Image save requested but accepted type not "
                "specified.\n"
                f"Accepted image formats include {valid_img_formats}\n"
                "Saving as .png"
            ),
            UserWarning,
        )
        ext = ".png"
    plotly_io.write_image(fig=fig, file=os.path.normpath(main + ext))
```

r-leyshon: I suppose what I'm saying is don't give the user any choice in the matter. They can tell you where to write the file and what format you want. The you just serve them up the correct filename and file extension(s) for what they asked for. They can go ahead and manually rename that but there's less room for things to go wrong. A bit like htmlcov behaviour.

cbrowning04: So what you're suggesting is to essentially just remove the ability for the user to name the file, just allow them to specify a parent dir and create the filename ourselves. Also I would allow them to choose a filetype, however if they choose an incorrect file type, throw an error and make them select a correct one rather than imputing it,

r-leyshon: does that sound OK? Should be able to securely test that approach as the warnings stuff is a bit janky, though we do need to lean into that problem at some time, now may not be the ideal time.

src/transport_performance/utils/defence.py
r-leyshon: Tests pushed. Resolving this thread.

tests/gtfs/test_validation.py
cbrowning04: @r-leyshon A PR has now been opened for this refactor. It gives the user less room for error while also raising errors if invalid file types are passed, rather than warnings

src/transport_performance/utils/defence.py
```diff
@@ -206,3 +223,30 @@ def _check_list(ls, param_nm, check_elements=True, exp_type=str):
...
+    None
+
+    Raises
+    ------
+    IndexError
+        Raises an error if the column (column_name) does not exist in the
+        dataframe
+
+    """
+    if column_name not in df.columns:
```
r-leyshon, line 249: @CBROWN-ONS this line is not covered in the test suite. This should be a straight forward test though.

src/transport_performance/gtfs/report/report_utils.py
```diff
@@ -0,0 +1,183 @@
...
+            "Report already exists at path: "
+            f"[{path}]."
+            "Consider setting overwrite=True"
+            "if you'd like to overwrite this."
+        )
+
+    # make gtfs_report dir
+    try:
+        os.mkdir(f"{path}/gtfs_report")
+    except FileExistsError:
```
r-leyshon, line 179: This line isn't covered by the test suite. Would this be a simple job to cover?

src/transport_performance/gtfs/validation.py
```diff
@@ -601,3 +655,803 @@ def get_route_modes(self):
...
+                    original_rows=rows,
+                )
+                base_columns = [
+                    item
+                    for item in list(filtered_tbl.columns)
+                    if item not in join_vars
+                ]
+                duplicate_counts = {}
+                for col in base_columns:
+                    duplicate_counts[col] = impacted_rows[
```
r-leyshon, line 1030: Currently no tests for this behaviour. How challenging to produce this? Would we need another fixture or does the Newport fixture present a repeated pair?

src/transport_performance/gtfs/validation.py
```diff
@@ -601,3 +655,803 @@ def get_route_modes(self):
...
+                    <div style="display: block;">
+                            <dd style="font-weight: bold;
+                                margin-inline-start: 0;
+                                display: inline-block;
+                                min-width: 160px;">{var}: </dd>
+                            <dt style="display: inline-block;">
+                            {duplicate_counts[var]}</dt>
+                    </div>"""
+                    )
+                table_html = table_html + "</div>"
```
r-leyshon, line 1073: The last block with no coverage in this diff. Would this need a check of the table HTML structure? If it's challenging to hit because of the logic above, it could be an idea to move this to a separate function where it would be simpler to feed in dummy `duplicate_counts` objects.

src/transport_performance/utils/defence.py
cbrowning04: This is a very straightforward test to implement. It did originally exist but I accidentally removed it. Should be easy to add

src/transport_performance/gtfs/report/report_utils.py
cbrowning04: I think this line would be covered if we simply ran through these lines once the gtfs_report had been created through testing

src/transport_performance/gtfs/validation.py
cbrowning04: We would need to add another feature as the Newport data does not include a repeated pair.

src/transport_performance/gtfs/validation.py
cbrowning04: As far as I can tell, there is no coverage as there are no repeated pairs in the test fixture.

src/transport_performance/gtfs/report/report_utils.py
cbrowning04: Hi @r-leyshon. I've pushed a commit that addresses this lack of coverage

src/transport_performance/utils/defence.py
cbrowning04: Hi @r-leyshon I've pushed a commit that addresses this lack of coverage

src/transport_performance/gtfs/gtfs_utils.py
```diff
@@ -60,3 +63,88 @@ def bbox_filter_gtfs(
     print(f"Filtered feed written to {out_pth}.")
 
     return None
+
+
+# NOTE: Possibly move to a more generalised utils file
+def convert_pandas_to_plotly(
+    df: pd.DataFrame, return_html=False, scheme: str = "dsc"
```
r-leyshon: Currently, there are no alternative schemes, providing this parameter potentially misleading. Suggest removing parameter and associated logic and keeping the hardcoded `schemes` variable.

r-leyshon: `return_html=False` needs type setting

src/transport_performance/gtfs/gtfs_utils.py
```diff
@@ -60,3 +63,88 @@ def bbox_filter_gtfs(
...
+    _type_defence(return_html, "return_html", bool)
+    if scheme not in list(schemes.keys()):
+        raise LookupError(
+            f"{scheme} is not a valid colour scheme."
+            f"Valid colour schemes include {list(schemes.keys())}"
+        )
+    if isinstance(df.columns, pd.MultiIndex):
+        raise TypeError(
+            "Pandas dataframe must have a single index," "not MultiIndex"
+        )
```
r-leyshon: Passes on df with multi index but raises when columns are multi index. Is this the behaviour you need? If so, make clear in TypeError.

Possibly raise when either index or columns are multi?

src/transport_performance/gtfs/report/report_utils.py
```diff
@@ -0,0 +1,183 @@
+"""Utils to assist in the creation of a HTML report for GTFS."""
```
r-leyshon, line 1: I expect the methods for this class to be internal only. I would have thought users would want the reports. But we can relax the testing for internals. Calling things like `TemplateHTML.insert()` would be for development only, right?

src/transport_performance/gtfs/report/report_utils.py
```diff
@@ -0,0 +1,183 @@
...
+        "block_id",
+        "shape_id",
+        "wheelchair_accessible",
+    ],
+    "shapes": [],
+}
+
+
+class TemplateHTML:
+    """A class for inserting HTML string into a docstring.
```
r-leyshon: Should this read:
`inserting HTML string into a template` instead of docstring?

src/transport_performance/gtfs/report/report_utils.py
```diff
@@ -0,0 +1,183 @@
...
+            The file path of the html template
+
+        Returns
+        -------
+        None
+
+        """
+        _handle_path_like(path, "path")
+        with open(path, "r", encoding="utf8") as f:
+            self.template = f.read()
```
r-leyshon, line 80: good idea to `f.close()` when done

src/transport_performance/gtfs/report/report_utils.py
```diff
@@ -0,0 +1,183 @@
...
+        return None
+
+    def insert(
+        self, placeholder: str, value: str, replace_multiple: bool = False
```
r-leyshon, line 84: Do we need the `replace_mutiple` parameter? The current behaviour raises if multiple are found. The alternative would be to leave placeholders in the HTML, which seems incorrect (unless I’ve misunderstood). Potentially we can drop this param and associated logic, what do you think?

r-leyshon: If decided to retain `replace_multiple`, the error message isn't formatted correctly when triggered. Suggest refactoring to:
```
"`replace multiple` requires True as found \n"
"multiple placeholder matches in template."
```

src/transport_performance/gtfs/report/report_utils.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def get_template(self) -> str:
+        """Get the template attribute of the TemplateHTML object.
+
+        Returns
+        -------
+        str
+            The template attribute
+
+        """
```
r-leyshon: Looks like the only reason we need this is to play nicely with pytest fixtures? Would be worth putting this reason within the docstring and ensuring this method is denoted as internal.

src/transport_performance/gtfs/report/report_utils.py
```diff
@@ -0,0 +1,183 @@
...
+        return self.template
+
+
+def set_up_report_dir(
```
r-leyshon: Please can you debug this:

```
set_up_report_dir("TESTING", overwrite=True)
  File "<stdin>", line 33, in set_up_report_dir
FileNotFoundError: [Errno 2] No such file or directory: 'TESTING/gtfs_report'
>>> set_up_report_dir("TESTING/TESTING1", overwrite=True)
Creating parent directory: /Users/richardleyshon/Documents/transport-network-performance/TESTING
  File "<stdin>", line 33, in set_up_report_dir
FileNotFoundError: [Errno 2] No such file or directory: 'TESTING/TESTING1/gtfs_report' 

```

It looks like `_check_parent_dir_exists` doesn't have the behaviour you need here. Needs adapting.

tests/gtfs/report/test_report_utils.py
```diff
@@ -0,0 +1,105 @@
...
+            placeholder="test_placeholder",
+            value="test_value",
+            replace_multiple=True,
+        )
+        assert (
+            expected_template == template_fixture.get_template()
+        ), "Test placeholder replacement not acting as expected"
+
+
+class TestSetUpReportDir(object):
```
r-leyshon, line 65: Tests in this class interact with test/data/gtfs/report . Best practice to avoid writing out lint on running test suite. Instead, use `pytest.tmpdir` fixture to create the temp dir structure you want to work with. Once the test run completes, any tmps will be cleaned up. An example implementation is in tests`/gtfs/test_gtfs_utils.py test_bbox_filter_gtfs_writes_as_expected`

tests/gtfs/report/test_report_utils.py
```diff
@@ -0,0 +1,105 @@
...
+                    "Report already exists at path: "
+                    "[tests/data/gtfs/report]."
+                    "Consider setting overwrite=True"
+                    "if you'd like to overwrite this."
+                )
+            ),
+        ):
+            set_up_report_dir("tests/data/gtfs/report")
+
+    def test_set_up_report_dir_on_pass(self, tmp_path):
```
r-leyshon: Suggest additional checks for the cases discovered above:
`set_up_report_dir("TESTING", overwrite=True)` `set_up_report_dir("TESTING/TESTING1", overwrite=True)`

tests/gtfs/report/test_report_utils.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def test_init(self, template_fixture):
+        """Test initialising the TemplateHTML class."""
+        expected_template = """<!DOCTYPE html>
+<html lang="en">
+
+<body>
+    <div>[test_placeholder] Tester [test_placeholder]</div>
+</body>
+"""
```
r-leyshon, line 36: Indentation shenanigans. A formatting issue with line breaks leading to a test failure if this is indented as expected.  The test is brittle to formatting changes. Potentially `str.replace(r"\n", "")` both templates (the fixture and the predicate) and then do the assertion?

tests/gtfs/report/test_report_utils.py
```diff
@@ -0,0 +1,105 @@
...
+        template_fixture.insert(
+            placeholder="test_placeholder",
+            value="test_value",
+            replace_multiple=True,
+        )
+        assert (
+            expected_template == template_fixture.get_template()
+        ), "Test placeholder replacement not acting as expected"
```
r-leyshon: Formatting issue once more. Rather than using an expected template, could we `assert "test_value" in template_fixture.get_template()`

tests/gtfs/test_gtfs_utils.py
```diff
@@ -40,3 +47,57 @@ def test_bbox_filter_gtfs_writes_as_expected(self, tmpdir):
...
+
+    def test_convert_pandas_to_plotly_on_pass(self):
+        """Test convert_pandas_to_plotly() when defences pass."""
+        test_df = pd.DataFrame(
+            {
+                "ID": [1, 2, 3, 4, 1],
+                "score": [45, 34, 23, 12, 23],
+                "grade": ["A", "B", "C", "D", "C"],
+            }
+        )
```
r-leyshon: As this is repeating code from above, a locally scoped fixture will do the trick here. Like how you handled the fixture in tests/gtfs/report/test_report_utils.py `@pytest.fixture(scope="function")`

src/transport_performance/gtfs/report/report_utils.py
cbrowning04: using `with` automatically closes the file once the `with` block is escaped. Happy to add `f.close()` however if you fell like it would make the code more readable and protect against the off chance the `with` block acts incorrectly.  
@r-leyshon

src/transport_performance/gtfs/report/report_utils.py
cbrowning04: The issue here was that the `path` param was being passed to `_check_parent_dir_exists()`, which would only create the directory up to `testing/testing` in the case that `path=testing/testing/testing` as it is only trying to create the 'parent' directory. Functionality has been updated to reflect this.

src/transport_performance/gtfs/validation.py
```diff
@@ -116,6 +131,20 @@ def __init__(
             raise ValueError(f"`units` accepts metric only. Found: {units}")
 
         self.feed = gk.read_feed(gtfs_pth, dist_units=units)
+        self.gtfs_path = gtfs_pth
+
+    def get_gtfs_files(self) -> list:
```
r-leyshon, line 167: I can think of 2 other conditions we would want to check for raise:

1. Fewer files than expected
2. Additional files than expected (perhaps a warning is more appropriate here if we can test it without having it show up in test reports)

As logic not implemented, this can be a feature request. Would you agree?

src/transport_performance/gtfs/validation.py
```diff
@@ -567,8 +609,20 @@ def summarise_routes(
...
+        day_route_count.columns = day_route_count.columns = [
+            "_".join(value) if "" not in value else "".join(value)
+            for value in day_route_count.columns.values
+        ]
+        day_route_count.columns = [
+            column.replace("amin", "min").replace("amax", "max")
+            for column in day_route_count.columns.values
+        ]
+
+        self.daily_route_summary = day_route_count.copy()
```
r-leyshon: Repeats logic in lines 530 to 544. Helper function? It can be an internal with a bit of light unit testing.

src/transport_performance/gtfs/validation.py
cbrowning04: Yes I agree with this. From my understanding you're referencing the actual gtfs files e.g., calendar.txt. I'm happy if we want to open this as a feature request, it would be a nice layer of protection to have.  
`gtfs-kit` may already have this functionality which we would need to confirm or deny first.

src/transport_performance/gtfs/validation.py
```diff
@@ -601,3 +655,803 @@ def get_route_modes(self):
...
+        _type_defence(target_column, "target_column", str)
+        _type_defence(plotly_kwargs, "plotly_kwargs", dict)
+        _type_defence(return_html, "return_html", bool)
+        _type_defence(width, "width", int)
+        _type_defence(height, "height", int)
+        _type_defence(xlabel, "xlabel", (str, type(None)))
+        _type_defence(ylabel, "ylabel", (str, type(None)))
+        _type_defence(save_html, "save_html", bool)
+        _type_defence(save_image, "save_iamge", bool)
+        _type_defence(img_type, "img_type", str)
```
r-leyshon: Consider whether a for loop will suit your needs in fewer lines:
```
for i in [day_column, target_column, xlabel, ylabel, img_type]:
  _type_defence(i, "grouped plot str inputs", (str, type(None)))
```

src/transport_performance/gtfs/validation.py
```diff
@@ -601,3 +655,803 @@ def get_route_modes(self):
...
+            )
+
+        # dataframe column defences
+        _check_column_in_df(df=summary_df, column_name=target_column)
+        _check_column_in_df(df=summary_df, column_name=day_column)
+
+        # convert column type for better graph plotting
+        summary_df["route_type"] = summary_df["route_type"].astype("object")
+
+        orientation = orientation.lower()
```
r-leyshon: second time lowering orientation. Good practise to do this but generally safer to get it done at the top of the module so you can rely on that behaviour downstream (and only have to do it the once!)

src/transport_performance/gtfs/validation.py
```diff
@@ -601,3 +655,803 @@ def get_route_modes(self):
...
+            xlabel=xlabel,
+            ylabel=ylabel,
+            save_html=save_html,
+            out_dir=out_dir,
+            save_image=save_image,
+            img_type=img_type,
+        )
+        return plot
+
+    def plot_trip_summary(
```
r-leyshon: I feel we can generalise plot summary to cope with the behavioural requirements for plot_route_summary and plot_trip_summary. Their behaviours are consistent for different dataframes, right? This would save us approx 300 lines in this module and more again in testing.

the suggestion is to remove plot_trip_summary and plot_route_summary in favour of a plot_summary function that can handle both of their needs.

src/transport_performance/gtfs/validation.py
```diff
@@ -601,3 +655,803 @@ def get_route_modes(self):
...
+            "agency": self.feed.agency,
+            "routes": self.feed.routes,
+            "stop_times": self.feed.stop_times,
+            "stops": self.feed.stops,
+            "trips": self.feed.trips,
+            "calendar": self.feed.calendar,
+        }
+
+        # determine which errors/warnings have rows that can be located
+        validation_table = self.is_valid()
```
r-leyshon, line 990: This bit merits some thinking about. I think we need a separate issue to discuss what to do with the module as a whole, now we have some more concrete outputs - the GTFS html reports. It’s likely that we can move some of the initial methods to internals as we expect the main use of this class to be `html_report()` . Will spin up a separate issue to capture this refactoring consideration.

src/transport_performance/gtfs/validation.py
```diff
@@ -601,3 +655,803 @@ def get_route_modes(self):
...
+                if len(rows) > 1
+                else "Unavailable"
+                for href, rows in zip(info_href, validation_dataframe["rows"])
+            ]
+
+        eval_temp = TemplateHTML(
+            path=(
+                "src/transport_performance/gtfs/report/"
+                "html_templates/evaluation_template.html"
+            )
```
r-leyshon, line 1162: From a functional purity perspective, we should pass this path in as a value to the method. Danger being that breaking that external dependency doesn’t get handled at instantiation / in defensive checks.

src/transport_performance/gtfs/validation.py
```diff
@@ -601,3 +655,803 @@ def get_route_modes(self):
...
+            out_pth=pathlib.Path(f"{report_dir}/gtfs_report/convex_hull.html"),
+            geoms="hull",
+            geom_crs=27700,
+        )
+        stops_temp = TemplateHTML(
+            (
+                "src/transport_performance/gtfs/report/"
+                "html_templates/stops_template.html"
+            )
+        )
```
r-leyshon, line 1253: another functional purity thing - lets pass it in as a parameter. No problem with these paths being set as default in the method signature though.

src/transport_performance/gtfs/validation.py
```diff
@@ -601,3 +655,803 @@ def get_route_modes(self):
...
+        summ_temp.insert("plotly_placeholder_2", trip_html)
+        summ_temp.insert(
+            "plotly_title_2",
+            f"Trip Summary by Day and Route Type ({summary_type})",
+        )
+        summ_temp.insert("date", date)
+        with open(
+            f"{report_dir}/gtfs_report/summaries.html", "w", encoding="utf8"
+        ) as summ_f:
+            summ_f.writelines(summ_temp.get_template())
```
r-leyshon: Good idea to `summ_f.close()`

tests/gtfs/test_validation.py
```diff
@@ -580,3 +602,203 @@ def test_summarise_routes_on_pass(self, gtfs_fixture):
...
+        )
+
+        # general save test
+        save_dir = os.listdir(os.path.join(tmp_path, "save_test"))
+        counts = {"html": 0, "png": 0}
+        for pth in save_dir:
+            if ".html" in pth:
+                counts["html"] += 1
+            elif ".png" in pth:
+                counts["png"] += 1
```
r-leyshon, line 744: In cases where we need to add more test conditions, these numbers would need fluffing. Not a push to change or remove, but worth noting. Feel free to disagree.

tests/gtfs/test_validation.py
```diff
@@ -580,3 +602,203 @@ def test_summarise_routes_on_pass(self, gtfs_fixture):
...
+                f"include {valid_img_formats}"
+            ),
+        ):
+            gtfs_fixture._plot_summary(
+                gtfs_fixture.daily_route_summary,
+                "route_count_mean",
+                save_image=True,
+                out_dir=os.path.join(tmp_path, "outputs"),
+                img_type="test",
+            )
```
r-leyshon: This looks like it belongs with `test__plot_summary_defences` not `test__plot_summary_on_pass`

tests/gtfs/test_validation.py
```diff
@@ -580,3 +602,203 @@ def test_summarise_routes_on_pass(self, gtfs_fixture):
...
+            gtfs_fixture.plot_route_summary(target_summary="mean"),
+            PlotlyFigure,
+        ), "plot_route_summary() failed to return plotly figure"
+
+    def test__plot_trip_summary_on_pass(self, gtfs_fixture):
+        """Test plot_trip_summary() calls _plot_summary() succesfully."""
+        gtfs_fixture.summarise_trips()
+        assert isinstance(
+            gtfs_fixture.plot_trip_summary(target_summary="mean"), PlotlyFigure
+        ), "plot_trip_summary() failed to return plotly figure"
```
r-leyshon: These tests are clones of each other, testing the same logic in the src code that I suggested would be better refactored to one method.

src/transport_performance/gtfs/validation.py
cbrowning04: **wont-fix**
Rationale: It takes approximately 3 lines (minimum) to check for 3 params using a loop which is the same as without a loop. This is because to maintain the correct error messages, a dictionary needs to be looped through.
There are no more than 3 type checks of one datatype, meaning using loops isn't beneficial.

tests/gtfs/test_validation.py
cbrowning04: Probably a wont-fix until we identify more test conditions

src/transport_performance/gtfs/validation.py
r-leyshon: Agreed to raise as issue, can I leave this with you?

src/transport_performance/gtfs/validation.py
r-leyshon: typo above, needs a new fixture that can reproduce the repeated pair context. I will raise an issue for this.

src/transport_performance/gtfs/validation.py
r-leyshon: agreed, closing out as linked to #133

src/transport_performance/gtfs/validation.py
r-leyshon: On reflection I agree with your position on this - not wishing to parameterise it incase we encourage non-compatible alternative templates.

src/transport_performance/gtfs/validation.py
r-leyshon: same as above.

tests/data/gtfs/report/gtfs_report/styles.css
r-leyshon, line 1: @CBROWN-ONS this file is lint left over from running the test suite, right? We good to remove this from the repo?

cbrowning04: Yes this file can be removed. It was previously used in testing but I decided not to use it at a later time.

src/transport_performance/gtfs/report/report_utils.py
```diff
@@ -0,0 +1,185 @@
...
+    ----------
+    template : str
+        A string containing the HTML template.
+
+    Methods
+    -------
+    insert(placeholder: str, value: str, replace_multiple: bool = False)
+        Insert values into the HTML template
+    get_template()
+        Returns the template attribute
```
r-leyshon: @CBROWN-ONS please update

r-leyshon: just patched this

testing/testing/testing/gtfs_report/styles.css
r-leyshon, line 1: @CBROWN-ONS  This issue with the lint being written out to directory. This file has been mistakenly committed and the test is writing out directories / css files (interestingly, the css file doesn't always appear though).
Uncertain what the cause is as your use of tmp_pth looks fine. May need to default to tmpdir if we can't get tmp_pth to behave as needed.

r-leyshon: just pushed up the removal of this lint and re-running. Problem seems to have gone away for me. If you could confirm on your end? Just run the pytest (maybe coverage run -m pytest too just to be sure) and confirm that no lint is being written out.

cbrowning04: @r-leyshon. It has now been removed from the branch completely. I can confirm that it isn't present and that it also does not impact any of the tests (the test suite is running smoothly).  
I think it was a remnant of an older test I had implemented that had since been removed.  
All seems fine on my end now

r-leyshon (approved):
All conversations resolved. New issues arising from this PR:
#129 
#130 
#131 
#133 
#134

PR #117, opened 2023-09-04
109 develop windows os ci

### LINKED ISSUE #109: Develop windows OS CI

r-leyshon:
## Description

Updating CI workflow to account for developers working on specific modules on Windows.
Fixes #109

## Git Workflow

This branch has been brought up to date with the current state of `dev` branch.
## CI Strategy

This PR introduces a new CI workflow:

`.github/workflows/windows-build.yml`

This workflow installs dependencies, runs a stripped back set of tests.

Step: “Run Integration Tests” uses the following pytest command:
`pytest -m runinteg --runinteg --deselect tests/osm/ --deselect tests/population`

It is basically checking the scraping logic only.

Step:  “Test with pytest”  uses the following command:
`pytest --deselect tests/osm/ --deselect tests/population # run only tests with no osmosis or geos dependencies`

This deselects all tests within the osm & population test modules. These modules can be assumed to be incompatible with windows os and as such, we would not adopt a windows-compatible badge.

## Handling Paths

`transport_performance.utils.defence._is_path_like` has been refactored to `transport_performance.utils.defence._handle_path_like` and breaking changes across the package updated (I noticed it had been used in `urban_centres/raster_uc` module).

This utility passes on windows, Mac and ubuntu. It first uses `os.path.realpath` to sort out relative paths or paths containing symbolic links. It then returns a `pathlib.Path` object. It no longer raises on discovery of `\\` in paths and I do not expect that to pose a problem.

A regex pattern in `tests/gtfs/test_validation` needed to be updated so that it can handle windows or macos type ValueError messages containing file paths.

A test in `tests/utils/test_defence.py` was removed as no longer required. This test checked the behaviour of raising on condition where path contains “\\\”. **Please let me know if there is something I have not accounted for…**

## Dependencies

`pandas` version fixed before 2.1.0
`dev` merged `matplotlib` req of >=3.7.0.


.github/workflows/windows-build.yml
```diff
@@ -0,0 +1,28 @@
+# This workflow will install Windows dependencies, run tests and lint with Python 3.9
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python
+
+name: Python package on Windows
+on:
+  push:
+    branches: ["109-develop-windows-os-ci"]
```
SergioRec: Do we want to change this to 
```
on:
  push:
  pull_request:
    branches: [ "dev", "main" ]
```

Otherwise my understanding is that it won't do anything after this branch is gone.

SergioRec (commented):
Added issue #120 about how to handle backslashes in string paths in macOS.

All code and tests run fine locally, minor issue/question in CI yaml.

.github/workflows/windows-build.yml
ethan-moss: Good catch! This has been updated in the latest commit.

ethan-moss (approved):

PR #132, opened 2023-09-14
119 e2e example

### LINKED ISSUE #119: e2e example

ethan-moss:
## Description

An initial end-to-end example. Closes #119. 

Changes in the PR are limited to just a new notebook and TOML file, and a slight change to `requirements.txt` to fix the `r5py` version.

## Motivation and Context

This adds in an initial attempt at developing an end-to-end run through, applying all the package modules developed so far. It also a first attempt at `analyse_network` and `metrics` stages. The motivation for merging this PR is so these attempts are stored with the main codebase to support further development at a later stage. It also fixes the version of `r5py` to the latest release (v0.1.0).

## Type of change

## How Has This Been Tested?

Manual QA-ing performed throughout, but not unit test developed for this PR. It is purely research inside a notebook.

Test configuration details:
* OS: macOS
* Python version: 3.9.12
* Java version: 11

## Advice for reviewer

You will need to re-install `requirements.txt` to review this PR, since we up-version `r5py` to 0.1.0.

A detailed review of the notebook is not required here, since it's not productionised code. Only a review of the overall application/methods plus attempting to re-run the notebook is required at this stage.

Input data sources have been shared separately.

## Checklist:

- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

## Additional comments

None

### DIFF
--- a/notebooks/e2e/e2e.py
+++ b/notebooks/e2e/e2e.py
@@ -26,9 +26,11 @@ import datetime
 import geopandas as gpd
 import pandas as pd
 import gtfs_kit as gk
+import folium
 
 from pyprojroot import here
 from shapely.geometry import box
+from folium.map import Icon
 from r5py import (
     TransportNetwork,
     TravelTimeMatrixComputer,
@@ -309,12 +311,11 @@ trans_net = TransportNetwork(
 travel_time_matrix_computer = TravelTimeMatrixComputer(
     trans_net,
     origins=centroid_gdf,
-    # destinations=centroid_gdf[centroid_gdf.within_urban_centre],
+    destinations=centroid_gdf[centroid_gdf.within_urban_centre],
     departure=datetime.datetime(2023, 8, 8, 8, 00),
-    transport_modes=[
-        TransportMode.TRANSIT,
-        TransportMode.WALK,
-    ],
+    departure_time_window=datetime.timedelta(hours=1),
+    max_time=datetime.timedelta(minutes=45),
+    transport_modes=[TransportMode.TRANSIT],
 )
 
 # %%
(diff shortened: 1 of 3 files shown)


notebooks/e2e/e2e.py
```diff
@@ -0,0 +1,921 @@
...
+            os.makedirs(dir_name)
+        m.save(save)
+
+    return m
+
+
+# %%
+# visualise for an ID within the UC
+UC_ID = 4110
+assert UC_ID in travel_times.to_id.unique()
```
r-leyshon, line 551: Use of assert outside of unit tests to be discouraged. See [bandit docs](https://bandit.readthedocs.io/en/latest/plugins/b101_assert_used.html) for more on that. OK with it being in a notebook but wouldn't want to see this in src.

notebooks/e2e/e2e.py
```diff
@@ -0,0 +1,921 @@
...
+# %%
+# merge the population input raster files to form one larger area
+# use regex to ensure 2020 data, CRS, and resolution are as expected in name
+if pop_config["override"]:
+    merge_raster_files(
+        here(pop_config["input_dir"]),
+        os.path.dirname(here(pop_config["merged_path"])),
+        os.path.basename(pop_config["merged_path"]),
+        subset_regex="GHS_POP_E2020_GLOBE_R2023A_54009_100_",
+    )
```
r-leyshon, line 149: `merge_raster_files` uses `check_parent_dir_exists(create=False)` - change to True or expose this parameter in the api for `merge_raster_files`, running in the e2e notebook required me to create the directory. Suggest raising separate issue.

notebooks/e2e/e2e.py
```diff
@@ -0,0 +1,921 @@
...
+    column="travel_time",
+    column_control_name="Travel Time",
+    point_control_name="Destination Centroid",
+    caption="Median Travel Time (mins)",
+    save=save_path,
+)
+
+# %%
+# visualise for an ID within the UC
+UC_ID = 3974
```
r-leyshon, line 598: Scenario to handle? I worked through the notebook with GHS pop2030 . I didn't get UC_ID 3974 for some reason.

`travel_times.to_id.unique()`
`…3945, 3946, 3947, 3948, 3949, 3950, 4004, 4005, 4006, 4007, 4008,…`

I have the intermediate outputs on my local machine if any of this would be useful in troubleshooting. Are the IDs consistent in time series? Is it a problem if they're not?

r-leyshon (changes requested):
All suggested adaptations are in #145 for your consideration @ethan-moss .

I popped a few comments in the e2e.py about things you may wish to discuss, but would not prevent a merge to dev. 

Overall, an excellent implementation of everyone's work over the last few months. The commenting was thorough enough for me to be able to find all the data sources needed to execute the cells. 

Feature request - data ingestion helpers for GHS & OSM? Happy to dicsuss, but wouldn't seem feasible for BODS.

notebooks/e2e/e2e.py
ethan-moss: Setting `create=True` here is not the right strategy, it will lead to incorrect and unintuitive function behaviour.

This function expects input files within an existing directory, and it's this directory we're checking. Creating the directory if it doesn't exist wouldn't work here because if the directory doesn't exist the input files do not exist either. Then, creating a new directory would lead to an error a few lines later when the search for .tif files happened - it's a new empty folder with no input data and so it would raise an error. It could also lead to lint if a user forgets to delete the empty folder. 

I'd recommend to not raise the separate issue. I'm happy to resolve this conversation if you are too?

This function has however been fully brought up to date with `defence.py` in PR #146.

notebooks/e2e/e2e.py
ethan-moss: This is an interesting find!

I don't think it's a scenario to handle since the IDs are purely supposed to be a unique identifier within the context of the area being explored. As long as the IDs are unique across the board, then that's all good.

The only thing I can think of that would cause changes in IDs would be the boundaries or thresholds are changed when reading in the population data. The IDs are assigned in 'order' from the top leftmost corner, so if any of the boundaries or thresholds are changed the IDs would change too. If it's not this, we should for sure explore in more detail to find the cause, since for the same boundary and thresholds the IDs should be the same.

notebooks/e2e/e2e.py
ethan-moss: Interesting read, I didn't know this, thanks! I agree with your view there - keeping an assert here is OK, we're not looking to compile to optimised bytecode so there is no chance for that assert to be removed. Even if it was, it's just a notebook and the worst case scenario is a KeyError would be raised a little later on.

notebooks/e2e/e2e.py
r-leyshon: I take your point here. Unfortunately, running the notebook results in `FileNotFoundError: Parent directory ~/Users/richardleyshon/Documents/transport-network-performance/data/external/population ~ not found on disk.`

If you would prefer not to expose the parameter, then it would be best to ensure the directory can be depended upon to exist. Chucking a .gitkeep in the appropriate folder and updating the .gitignore behaviour, for example.

ethan-moss: This is the expected behaviour - the error you are getting is the exact error that needs to be thrown. If the input directory doesn't exist then an error is needed since it can't proceed without any input data.

I wouldn't be in favour of a solution that has a fixed requirement on the expected input directory either. That solution would be too rigid, and wouldn't work well when installing the package from pypi say (can't enforce the github repo structure).

The current implementation is the correct one for me i.e. no input directory = error - the function needs input data to work (the function is globbing files within the directory, so it's the directory's existence that needs to be checked first).

Let me know if I'm missing anything.

Issue #58, opened 2023-08-09
utils.raster coverage


r-leyshon:
Please reference my comments within merged [17 pop merge resample](https://github.com/datasciencecampus/transport-network-performance/pull/30/files#diff-c35a6e75869dd71399741e51bc30426584029ce71b1aa29f9dd6cacbc1d603bfR163-R165) . I've highlighted where the coverage gaps are and the sort of strategy I would suggest. Happy to help out with this if required.

              Would be worth seeing if [any of the utility functions I’ve written](https://github.com/datasciencecampus/transport-network-performance/blob/9cb601de0461002f71c39e6a79830db70d4b4131/src/transport_performance/utils/defence.py) (not yet merged to main) would suffice. Or refactoring them. Example implementation below:

```
from transport_performance.utils.defence import _check_parent_dir_exists
…
```
Then in `raster.merge_raster_files`:
```
_check_parent_dir_exists(pth=output_filepath, param_nm=“output_filepath”, create=True)
```
If this meets your behavioural reqs, this function has already been tested so you should be good to go.

_Originally posted by @r-leyshon in https://github.com/datasciencecampus/transport-network-performance/pull/30#discussion_r1288715714_

PR #138, opened 2023-09-18
121 resample file directory

### LINKED ISSUE #121: `sum_resample_file()` does not correctly create a new directory
**Describe the bug**

`sum_resample_file()` does not correctly create a new directory if one does not already exist.

**Environment**

 - OS: macOS
 - Python version: 3.9.12
 - Java version: 11

**Steps to reproduce**

1. Call `sum_resample_file()` creating an output in a new sub directory.
2. The following error message is raised:
```
FileNotFoundError: [Errno 2] No such file or directory: ...
```

**Expected behavior**

New directory is build and the file is saved within it

**Screenshots**

None

**Test data/code**

None

**Additional context**

This can be fixed by editing line 165 to use `os.path.dirname(output_filepath)` instead of the file path itself. May want to also consider using `os.makedirs()` instead of `os.mkdir`, so that new directories are created recursively.

https://github.com/datasciencecampus/transport-network-performance/blob/c41f12c8a9ed881fc1c7f7816ef7c04a80c65a03/src/transport_performance/utils/raster.py#L163-L165

Work around until issue is fixed:  create new sub-directory manually before calling function.

ethan-moss:
## Description

Updates to `transport_performance.utils.raster`:
 - Migrated `merge_raster_files()` and `sum_resample_file()` to use `defence.py` functions. Fixes #121

## Motivation and Context

Resolve #121: now `sum_resample_file()` writes to new directories if they don't already exist, using `_check_parent_dir_exists()`. 

Took this opportunity to update `merge_raster_files()` with `_check_parent_dir_exists()` at the same time.

## Type of change

## How Has This Been Tested?

`sum_resample_file()` unit test updated to save in a new sub-directory, to demonstrate fix + ensure it does not come back. Added an extra unit test to test failing `sum_resample_file()` inputs.

Also tested 'manually' using 'real' raster data: calling `sum_resample_file()` and saving the output to a directory that does not exist in advance. Saves to new directory as expected.

Will update `merge_raster_files()` unit tests as part of resolving issue #58.

Test configuration details:
* OS: macOS
* Python version: 3.9.12
* Java version: 11

## Checklist:

## Additional comments

None

### DIFF
--- a/src/transport_performance/utils/raster.py
+++ b/src/transport_performance/utils/raster.py
@@ -13,6 +13,10 @@ import rioxarray
 
 from rioxarray.merge import merge_arrays
 from rasterio.warp import Resampling
+from transport_performance.utils.defence import (
+    _check_parent_dir_exists,
+    _is_expected_filetype,
+)
 
 
 def merge_raster_files(
@@ -146,8 +150,7 @@ def sum_resample_file(
 
     """
     # defend against case where the provided input dir does not exist
-    if not os.path.exists(input_filepath):
-        raise FileNotFoundError(f"{input_filepath} can not be found")
+    _is_expected_filetype(input_filepath, "input_filepath", exp_ext=".tif")
 
     xds = rioxarray.open_rasterio(input_filepath, masked=True)
 
@@ -161,7 +164,6 @@ def sum_resample_file(
     )
 
     # make output_filepath's directory if it does not exist
-    if not os.path.exists(os.path.dirname(output_filepath)):
-        os.mkdir(output_filepath)
+    _check_parent_dir_exists(output_filepath, "output_filepath", create=True)
 
     xds_resampled.rio.to_raster(output_filepath)
--- a/src/transport_performance/utils/raster.py
+++ b/src/transport_performance/utils/raster.py
@@ -77,9 +77,9 @@ def merge_raster_files(
 
     """
     # defend against case where the provided input dir does not exist
-    # note: not able to use `_check_parent_dir_exists()` here since its a dir
-    if not os.path.exists(input_dir):
-        raise FileNotFoundError(f"{input_dir} can not be found")
+    # add a dummy file to purely check if provided parent directory exists
+    dummy_path = os.path.join(input_dir, "dummy.txt")
+    _check_parent_dir_exists(dummy_path, "input_dir", create=False)
 
     # get tif files in directory, ensure some exist and select subset via regex
     tif_filepaths = glob.glob(f"{input_dir}/*.tif")
--- a/src/transport_performance/utils/raster.py
+++ b/src/transport_performance/utils/raster.py
@@ -107,7 +107,11 @@ def merge_raster_files(
     xds_merged = merge_arrays(arrays)
 
     # create full filepath for merged tif file and write to disk
+    # check expected file type and parent dir exists
     MERGED_DIR = os.path.join(output_dir, output_filename)
+    _is_expected_filetype(
+        MERGED_DIR, "MERGED_DIR", check_existing=False, exp_ext=".tif"
+    )
     _check_parent_dir_exists(MERGED_DIR, "MERGED_DIR", create=True)
     xds_merged.rio.to_raster(MERGED_DIR)
 
--- a/src/transport_performance/utils/raster.py
+++ b/src/transport_performance/utils/raster.py
@@ -77,6 +77,7 @@ def merge_raster_files(
 
     """
     # defend against case where the provided input dir does not exist
+    # note: not able to use `_check_parent_dir_exists()` here since its a dir
     if not os.path.exists(input_dir):
         raise FileNotFoundError(f"{input_dir} can not be found")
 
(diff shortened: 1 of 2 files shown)


src/transport_performance/utils/raster.py
```diff
@@ -73,6 +77,7 @@ def merge_raster_files(
 
     """
     # defend against case where the provided input dir does not exist
+    # note: not able to use `_check_parent_dir_exists()` here since its a dir
```
cbrowning04: It is mentioned here that `_check_parent_dir_exists()` can not be used. In theory, a dummy file name e.g., `dummy.txt` can be appended onto `input_dir` to include this functionality.  
We would need to have `create=False` in this instance as we only want to know if the directory exists.  
While this isn't a required change, it would improve consistency across all directory checks.

src/transport_performance/utils/raster.py
```diff
@@ -13,6 +13,10 @@
 
 from rioxarray.merge import merge_arrays
 from rasterio.warp import Resampling
+from transport_performance.utils.defence import (
+    _check_parent_dir_exists,
+    _is_expected_filetype,
+)
 
 
 def merge_raster_files(
```
cbrowning04, line 22: On a successful run, this function prints the inputs and output. This could be considered obnoxious. Is this something we want to remain? Most likely this won't be changed, especially if it is part of the `rioxarray` package.  
image

src/transport_performance/utils/raster.py
```diff
@@ -101,12 +106,9 @@ def merge_raster_files(
...
     xds_merged = merge_arrays(arrays)
 
-    # make output_dir if it does not exist
-    if not os.path.exists(output_dir):
-        os.mkdir(output_dir)
-
     # create full filepath for merged tif file and write to disk
     MERGED_DIR = os.path.join(output_dir, output_filename)
+    _check_parent_dir_exists(MERGED_DIR, "MERGED_DIR", create=True)
     xds_merged.rio.to_raster(MERGED_DIR)
```
cbrowning04, line 116: I recommend using `_is_expected_filetype()` here on MERGED_DIR.  
When using an invalid file extension (e.g., .html or .test), a ValueError is raised by rioxarray in `xds_merged.rio.to_raster(MERGED_DIR)`. This would be better suited with the error message created using `_is_file_type_expected()`.  
Current raised error:  
image

tests/utils/test_raster.py
```diff
@@ -241,3 +273,40 @@ def test_sum_resample_file(self, resample_xarr_fpath: str) -> None:
         # assert correct resampling values (summing consitiuent grids)
         expected_result = np.array([[[14, 22], [46, 54]]])
         assert np.array_equal(expected_result, xds_out.to_numpy())
+
+    @pytest.mark.parametrize(
```
cbrowning04, line 277: Are we able to include the 'match' parameter of pytest.raises() here?  
This will ensure that the tests are detecting the correct errors, and not passing on other errors of the same type.

cbrowning04, line 277: I would also suggest to possibly include a test that looks for a directory that doesn't exist to accompany the tests that looks for a file that does not exist

cbrowning04 (changes requested):
Overall use of defence.py functions is successful throughout. All uses of the defency.py functions act as expected.  

Most suggestions have been made for improvements.  Most of these suggestions are QOL and consistency/reliability suggestions.  

One of these suggestions recommends improving error communication to the user as an unclear error is raised that may cause confusion and require the user to complete more research into the error than necessary.

src/transport_performance/utils/raster.py
ethan-moss: This printing is coming from the fact that you're using a Jupyter notebook and by default the cell prints out the returned value of the function if it's not assigned to a variable. `merge_raster_files()` returns the boundaries of all the inputs and the output so that a user could check their extents should they wish e.g., to make sure there are no gaps or overlaps in the merging.

Are you happy to resolve this conversation based on this?

src/transport_performance/utils/raster.py
ethan-moss: Yes, this could work - great idea! I'll make it very clear in a comment the origin of the dummy file, so that we don't forget going forward.

ethan-moss: I think this is now covered by commit e7ce76b. Happy to resolve this conversation if you agree.

src/transport_performance/utils/raster.py
ethan-moss: Great catch! I've fix this one in 90451bb. Happy to close this comment out if you agree.

tests/utils/test_raster.py
ethan-moss: Good idea. This one is now captured in a738242. Happy to close this conversation if you are too.

ethan-moss: I'm not sure this test case is a valid one. The input to this function is a file path, and the existence of the file is what's being check. So, if the directory doesn't exist then input file can't exist either. Nonetheless, it's a very quick test to do so I've added it in commit f71625b. Let me know what you think. I'm happy to resolve this conversation if you are too.

cbrowning04: I agree that it may not be needed. Never the less it's only a small test and a few lines so it wouldn't hurt to have. I'm happy to merge this PR with dev once the runners have passed. Thanks for the quick changes @ethan-moss !

cbrowning04 (approved):

PR #139, opened 2023-09-18
108 update  is expected filetype defence function to check multiple extensions

### LINKED ISSUE #108: Update `_is_expected_filetype()` defence function to check multiple extensions

Update `_is_expected_filetype()` (in `utils/defence.py`), to allow multiple extensions to be checked (instead of a single string). This can be useful when multiple extensions are acceptable (e.g. .tif, and .tiff). This relates to #86 where an updated version of this function could be used to close out this exact issue.

Could either accept a regex or maybe a list of acceptable extensions and check against each.

**Additional context**

None

r-leyshon:
## Description

Fixes #108 

## Motivation and Context

Make function flexible to single string filetype or list of acceptable string filtypes, eg [".tiff", ".tif"]

## Type of change

## How Has This Been Tested?

Coverage for this diff at 100%
Tested with pytest

Test configuration details:
* OS: macos

## Checklist:

### DIFF
--- a/src/transport_performance/utils/defence.py
+++ b/src/transport_performance/utils/defence.py
@@ -101,14 +101,15 @@ def _is_expected_filetype(pth, param_nm, check_existing=True, exp_ext=".zip"):
         The name of the parameter being tested. Helps with debugging.
     check_existing : bool
         Whether to check if the filetype file already exists. Defaults to True.
-    exp_ext: str
-        The expected filetype.
+    exp_ext: (str, list)
+        The expected filetype as a string, including the full stop. Or a list
+        of file extension strings to check.
 
     Raises
     ------
     TypeError: `pth` is not either of string or pathlib.PosixPath.
     FileExistsError: `pth` does not exist on disk.
-    ValueError: `pth` does not have the expected file extension.
+    ValueError: `pth` does not have the expected file extension(s).
 
     Returns
     -------
@@ -120,7 +121,12 @@ def _is_expected_filetype(pth, param_nm, check_existing=True, exp_ext=".zip"):
     _, ext = os.path.splitext(pth)
     if check_existing and not os.path.exists(pth):
         raise FileExistsError(f"{pth} not found on file.")
-    if ext != exp_ext:
+    if isinstance(exp_ext, list):
+        is_correct = ext in exp_ext
+    elif isinstance(exp_ext, str):
+        is_correct = ext == exp_ext
+
+    if not is_correct:
         raise ValueError(
             f"`{param_nm}` expected file extension {exp_ext}. Found {ext}"
         )
--- a/src/transport_performance/utils/defence.py
+++ b/src/transport_performance/utils/defence.py
@@ -4,6 +4,7 @@ import numpy as np
 import os
 from typing import Union
 import pandas as pd
+import warnings
 
 
 def _handle_path_like(pth, param_nm):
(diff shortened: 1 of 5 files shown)


src/transport_performance/utils/defence.py
```diff
@@ -120,7 +126,12 @@ def _is_expected_filetype(pth, param_nm, check_existing=True, exp_ext=".zip"):
     _, ext = os.path.splitext(pth)
     if check_existing and not os.path.exists(pth):
         raise FileExistsError(f"{pth} not found on file.")
```
cbrowning04: This isn't being tested within `tests.utils.test_defence.py::Test_IsExpectedFiletype`. This would be a useful test to have at the lowest level.

src/transport_performance/utils/defence.py
```diff
@@ -120,7 +126,12 @@ def _is_expected_filetype(pth, param_nm, check_existing=True, exp_ext=".zip"):
     _, ext = os.path.splitext(pth)
```
cbrowning04, line 135: It would be useful to have some sort of type checking on the parameters before this line.  
Suggested implementation:  
```python
_type_defence(param_nm, "param_nm", str)
_type_defence(check_existing, "check_existing", bool)
_type_defence(exp_ext, "exp_ext", (str, list))
```
It is useful to note that this implementation would require `_type_defence()`  to be moved above `_is_expected_filetype()`, which should be a fairly easy change.

src/transport_performance/utils/defence.py
```diff
@@ -120,7 +126,12 @@ def _is_expected_filetype(pth, param_nm, check_existing=True, exp_ext=".zip"):
...
     if check_existing and not os.path.exists(pth):
         raise FileExistsError(f"{pth} not found on file.")
-    if ext != exp_ext:
+    if isinstance(exp_ext, list):
+        is_correct = ext in exp_ext
+    elif isinstance(exp_ext, str):
+        is_correct = ext == exp_ext
```
cbrowning04: It could be worth lowering both `ext` and `exp_ext` here. Each item in `exp_ext` would need to be lowered in the case of a list. This could be a small QOL improvement

src/transport_performance/utils/defence.py
cbrowning04: This should be a `FileNotFoundError` rather than a `FileExistsError`. The latter indicates that the file is present, when it has been determined that it is not.

src/transport_performance/utils/defence.py
```diff
@@ -90,7 +90,12 @@ def _check_parent_dir_exists(
     return None
 
 
-def _is_expected_filetype(pth, param_nm, check_existing=True, exp_ext=".zip"):
+def _is_expected_filetype(
+    pth: Union[pathlib.Path, str],
+    param_nm: str,
+    check_existing: bool = True,
+    exp_ext: Union[str, list] = ".zip",
+) -> None:
```
cbrowning04, line 99: It is very likely that people pass file extensions to exp_ext without a leading '.' e.g., 'html' rather than '.html'. It may be useful to either detect and adjust this, or to mention is explicitly in the docstring

cbrowning04 (changes requested):
Functionality works great along with tests.  
Small changes requested in both the source code and the tests.

src/transport_performance/utils/defence.py
r-leyshon: Good spot! Will check asap

r-leyshon: Ah, my bad. Will patch now.

src/transport_performance/utils/defence.py
r-leyshon: Yeah did think about that. OK, I'll probably handle that case and put something in the docstring, as I'm very likely the one who will misuse it in that way,

r-leyshon:
@CBROWN-ONS we're good - test_raster has been refactored to assert for `FileNotFoundError` and all feedback implemented. Great work on spotting the error consistency!

cbrowning04 (approved):

PR #141, opened 2023-09-19
118 test  handle path like

### LINKED ISSUE #118: test _handle_path_like

Unit test for `utils.defence._handle_path_like` - behaves as expected on all reasonable path formats.

`pathlib.PureWindowsPath` [docs on pathlib](https://docs.python.org/3/library/pathlib.html#pathlib.PureWindowsPath) should be able to instantiate Windows-type paths on a unix-like. The same code should therefore be viable across all os.If this proves not to be possible, then suggest using an 'on-windows' / 'on-unix' flags to ensure the correct paths get tested on correct os.

r-leyshon:
## Description

Fixes #118 

## Motivation and Context

Implement tests for all sensible path parameters, regardless of operating system. Including passing of windows style path strings when working user is on macos. Including use of pyprojroot.here on windows and macos.

## Type of change

- [x] Refactored code & test suite

## How Has This Been Tested?

Tests pass on runners for windows, mac and ubuntu.
Assertions vary if on windows os - return type is WindowsPath with backward slashes. 

Test configuration details:
* OS: macos

## Checklist:

### DIFF
--- a/src/transport_performance/utils/defence.py
+++ b/src/transport_performance/utils/defence.py
@@ -13,7 +13,8 @@ def _handle_path_like(
     """Handle path-like parameter values.
 
     Checks a path for symlinks and relative paths. Converts to realpath &
-    outputs pathlib.PosixPath object (platform agnostic).
+    outputs pathlib.PosixPath or pathlib.WindowsPath object
+    (platform agnostic).
 
     Parameters
     ----------
--- a/src/transport_performance/utils/defence.py
+++ b/src/transport_performance/utils/defence.py
@@ -40,7 +40,8 @@ def _handle_path_like(
     pth_str = str(pth).replace("\\", "/")
 
     # Ensure returned path is not relative or contains symbolic links
-    pth = pathlib.Path(pth_str).resolve(strict=False)  # True checks exists
+    pth = os.path.realpath(pth_str)
+    pth = pathlib.Path(pth)
 
     return pth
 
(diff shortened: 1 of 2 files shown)


src/transport_performance/utils/defence.py
```diff
@@ -27,18 +30,20 @@ def _handle_path_like(pth, param_nm):
...
     -------
     pathlib.Path
-        Platform agnostic representation of pth.
+        Platform agnostic representation of pth. On unix-like a PosixPath is
+        returned. On windows a WindowsPath is returned. Both are children of
+        pathlib.Path.
 
     """
-    if not isinstance(pth, (str, pathlib.Path)):
+    if not isinstance(pth, (str, pathlib.Path, pathlib.PosixPath)):
```
cbrowning04: Am I correct in assuming this addition is redundant? This is because `pathlib.Path` is the parent class of `pathlib.PosixPath`, meaning that any `pathlib.PosixPath` passed is an instance of `pathlib.Path`

cbrowning04 (changes requested):
Hi @r-leyshon . Everything looks really good and works well. `_handle_path_like()` works as expected and handles various path formats (further reinforced by the tests).  
I have one comment which you may feel is a 'wontfix' as it's very small.  
Otherwise I'm happy to merge with dev.  
Very detailed tests!

src/transport_performance/utils/defence.py
r-leyshon: yes correct! I will patch now. Thanks again for your patience. Good to have these resolved.

cbrowning04 (approved):

PR #144, opened 2023-09-21
53 54 56 population defence

### LINKED ISSUE #53: saving plots to files in root
Applies to all backend types in `RasterPop.plot()`:

Logic breaks when user tries:
```
>>> self.plot(which="cartopy", save="foobar.png")
  File "<stdin>", line 150, in plot
  File "<stdin>", line 560, in _plot_cartopy
```
Os.mkdir doesn’t like cases where the parent directory returned equals “”.  [There is a utility that should do what you need](https://github.com/datasciencecampus/transport-network-performance/blob/9cb601de0461002f71c39e6a79830db70d4b4131/src/transport_performance/utils/defence.py#L31C1-L32C1), yet to be merged to dev though.

Example implementation:

```
from transport_performance.utils.defence import _check_parent_dir_exists

…

_check_parent_dir_exists(save, ”save”, create=True)
```
Agreed to raise as a technical debt issue.

_Originally posted by @r-leyshon in https://github.com/datasciencecampus/transport-network-performance/pull/44#discussion_r1288574431_

### LINKED ISSUE #54: defence utility check polygon
Occurs to me that this snippet would be a useful utility function. Agreed to raise as technical debt issue.

_Originally posted by @r-leyshon in https://github.com/datasciencecampus/transport-network-performance/pull/44#discussion_r1288578166_

### LINKED ISSUE #56: assertion failure messages where possible

Helpful in diagnosing the source of breaking changes and the expected behaviours.

**Additional context**

Agreed that this is not always possible in cases where `pd.testing` is used for example, frame assert equal doesn't seem to allow for this. But in cases of regular `assert` we should be including these.

ethan-moss:
## Description

Brings `population` module up to speed with `defence.py` functions.

Closes #54, Closes #56. Also indirectly fixes #53 - not original intention of PR, but bringing it up to speed with `defense.py` fixes the issue.

## Motivation and Context

`population` module was written before `defense.py` existed. The changes made in this PR aim to bring `population` inline with these defensive functions, to improve the `population` module. It also improves the unit testing, to assert the expected function argument types for public facing class methods.

## Type of change

## How Has This Been Tested?

All changes have been tested locally, to ensure the expected raises are triggered. New, passing, unit tests have also been added to reflect these changes.

Test configuration details:
* OS: macOS
* Python version: 3.9.13
* Java version: 11

## Advice for reviewer

A breaking change is being introduced into dev that will likely cause `population` unit tests to fail when merging to dev. This needs to be resolved before merging.

I would also appreciate your opinion of the regexs used within the `match` argument of `pytest.raises` calls here. I believe they are such that it's a good balance between capturing the key aspects, being robust to changes in raises statements, and not using the expected raise statement verbatum - which would make the tests very rigid and more difficult to maintain. If you think this compromise is not good, do say and I'm happy to change it. 

Note: type/input checking of arguments has only been added to public facing class methods, not internal methods or kwargs (type checking kwarg arguments would be subject to any external packages). If there are any 'key' kwarg variables you'd prefer to have type checked please say.

## Checklist:

> Note: this last point needs to be resolved before merging.

## Additional comments

None.

### DIFF
--- a/src/transport_performance/population/rasterpop.py
+++ b/src/transport_performance/population/rasterpop.py
@@ -20,6 +20,7 @@ from transport_performance.utils.defence import (
     _is_expected_filetype,
     _type_defence,
     _handle_path_like,
+    _check_parent_dir_exists,
 )
 
 
@@ -503,9 +504,7 @@ class RasterPop:
 
         # write to file if filepath is given
         if save is not None:
-            out_df = os.path.dirname(save)
-            if not os.path.exists(out_df):
-                os.mkdir(out_df)
+            _check_parent_dir_exists(save, "save", create=True)
             m.save(save)
             m = None
 
@@ -676,9 +675,7 @@ class RasterPop:
         # write to file if filepath is given, since there is no figure, need to
         # get the current figure and resize it to match the axis before saving
         if save is not None:
-            out_df = os.path.dirname(save)
-            if not os.path.exists(out_df):
-                os.mkdir(out_df)
+            _check_parent_dir_exists(save, "save", create=True)
             fig = plt.gcf()
             fig.set_size_inches(*figsize)
             fig.savefig(save)
@@ -717,9 +714,7 @@ class RasterPop:
 
         # write to file if filepath is given
         if save is not None:
-            out_df = os.path.dirname(save)
-            if not os.path.exists(out_df):
-                os.mkdir(out_df)
+            _check_parent_dir_exists(save, "save", create=True)
             fig.savefig(save)
             ax = None
 
--- a/src/transport_performance/population/rasterpop.py
+++ b/src/transport_performance/population/rasterpop.py
@@ -189,8 +189,9 @@ class RasterPop:
         - Cartopy backend: `help(RasterPop._plot_cartopy)`
 
         """
-        # input type defence
+        # input type defence - setting which to lower case to improve usability
         _type_defence(which, "which", str)
+        which = which.lower()
         if save is not None:
             save = _handle_path_like(save, "save")
 
--- a/src/transport_performance/population/rasterpop.py
+++ b/src/transport_performance/population/rasterpop.py
@@ -727,6 +727,9 @@ class RasterPop:
             writing to file, None is return.
 
         """
+        # handle kwarg type checks
+        _type_defence(figsize, "figsize", tuple)
+
         # handle matplotlib and rioxarry steps
         fig, ax = plt.subplots(figsize=figsize)
         self._xds.plot(ax=ax)
--- a/src/transport_performance/population/rasterpop.py
+++ b/src/transport_performance/population/rasterpop.py
@@ -125,6 +125,7 @@ class RasterPop:
         self._read_and_clip(aoi_bounds, aoi_crs, var_name)
 
         # round population estimates, if requested
+        _type_defence(round, "round", bool)
         if round:
             self._round_population()
 
--- a/src/transport_performance/population/rasterpop.py
+++ b/src/transport_performance/population/rasterpop.py
@@ -326,12 +326,9 @@ class RasterPop:
             When `urban_centre_bounds` is not a shapely Polygon.
 
         """
-        # defend against case where urban_centre_bounds is not a polygon
-        if not isinstance(urban_centre_bounds, Polygon):
-            raise TypeError(
-                f"Expected type {Polygon.__name__} for `urban_centre_bounds`, "
-                f"got {type(urban_centre_bounds).__name__}."
-            )
+        # input type defences
+        _type_defence(urban_centre_bounds, "urban_centre_bounds", Polygon)
+        _type_defence(urban_centre_crs, "urban_centre_crs", (str, type(None)))
 
         # match the crs if is one isn't provided - default assumption if this
         # arg is not set, and a CRS is needed when building the gdf below
--- a/src/transport_performance/population/rasterpop.py
+++ b/src/transport_performance/population/rasterpop.py
@@ -16,6 +16,9 @@ from typing import Union, Type, Tuple
 from shapely.geometry.polygon import Polygon
 from matplotlib import colormaps
 from cartopy.mpl.geoaxes import GeoAxes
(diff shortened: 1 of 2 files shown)


src/transport_performance/population/rasterpop.py
```diff
@@ -16,6 +16,12 @@
...
 from cartopy.mpl.geoaxes import GeoAxes
+from transport_performance.utils.defence import (
+    _is_expected_filetype,
+    _type_defence,
+    _handle_path_like,
+    _check_parent_dir_exists,
+)
 
 
 class RasterPop:
```
cbrowning04, line 27: ### Proposed Change
For the private plotting functions in `RasterPop`, type defences should be added.  
These functions include:  

- `._plot_folium()`
- `._plot_cartopy()`
- `._plot_matplotlib()`  

The rationale for this is that while they are private functions, kwargs can be passed to then through _.plot()_. Because of this, type defences should be included.  

### Recommendation: 
While new defences are being added, these do not need to be explicitly tested as the defensive functions are being tested in `tests/utils/test_defence.py`. You are more than welcome to test them in this function however if you feel it add better security.

src/transport_performance/population/rasterpop.py
```diff
@@ -261,10 +270,11 @@ def _round_population(self) -> None:
...
 
-    def _threshold_population(self, threshold: int) -> None:
+    def _threshold_population(self, threshold: Union[int, float]) -> None:
         """Threshold population data."""
         # `where()` is working differently to expectation here - it keeps
         # values above the threshold and otherwise sets them to nan.
+        _type_defence(threshold, "threshold", (int, float))
         self._xds = self._xds.where(self._xds >= threshold)
 
     def _to_geopandas(self, round: bool = False) -> None:
```
cbrowning04, line 281: ### Suggested Change
The `._to_geopndas()` function has no type checking for `round`.
I'm aware that there is a type defence for `round` in `.get_pop()`, however if `._to_geopndas()` was to be implemented elsewhere it would lack type defences.  

Another reason to include type defences here is to be more consistent with other private functions. All other private functions in `RasterPop`, more specifically the ones used by `.get_pop()` have type defences for their parameters within the function, with `_tp_geopandas()` being the exception.

### Implementation (OPTIONAL)
Insert the below snippet onto line 281 of `src/transport_performance/population/rasterpop.py`:  

```python
_type_defence(round,` "round", bool)
```

src/transport_performance/population/rasterpop.py
ethan-moss: Good point - I was toying with this and went on the side of maintainability. I was never fully convinced by my decision, so happy to do this. I'll work on it now since I have some time. In terms of unit tests, it's no so much about testing the methods of the functions inside `defence.py`, but more that unit tests are needed so that we can ensure the right types are being passed to the plotting functions, and that the correct errors are raised if it's not. Would you agree? If so, I'll add in respective unit tests too.

src/transport_performance/population/rasterpop.py
ethan-moss: Type checking for round does exist here:

https://github.com/datasciencecampus/transport-network-performance/blob/c70dabb0a84229db423ee1c11aefe75c88eed305/src/transport_performance/population/rasterpop.py#L129-L131

`_to_geopandas()` is an internal, private class method that has no intention of being used elsewhere (otherwise we should move it to a utils function). It also uses the type checked variable, so it is not possible for it to be run unchecked.

Happy to resolve this conversation if you are?

src/transport_performance/population/rasterpop.py
cbrowning04: Unit tests are entirely optional here in my opinion as there are tests for the defensive functions, meaning that we know they will work regardless. It's entirely up to you though!

src/transport_performance/population/rasterpop.py
cbrowning04: After reading this again I completely agree. I''ll resolve this now 😄

ethan-moss: OK great! Thanks @CBROWN-ONS! All helps to save a line :) If the `_to_geopandas()` method could be useful utils function do shout though.

src/transport_performance/population/rasterpop.py
ethan-moss: What are your thoughts on unit testing that the expected errors are raised if an incorrect type was passed as a kwarg? Not so much to test the `defence.py` function, but that the right type checking and error messages are being raised.

cbrowning04, line 27: ### Suggested Change  
_NOTE: This change is completely optional and doesn't fix any broken functionality_  

To improve usability, the `which` parameter could be lowered (`.lower()`) once it is passed to `.plot()`. This would disregard cases where the user uses the wrong capitalisation.  

This is an overall QOL improvement, but I'm happy for it to not be implemented if you feel that it is unnecessary.

ethan-moss: Great shout! I'll add this in.

ethan-moss: This has been implemented in commit 3f8dfac. There is a corresponding unit test too. Let me know if you consider this conversation closed.

cbrowning04: Resolving this conversation as it's complete from commit `3f8dfac`

ethan-moss: This should now be closed out by the commits f0ec246, 8b2700f, ca2cc5b and af1a0b0. It's made the diff much larger, but these are primarily pytest parameterised configs for the unit tests being formatted over multiple line.

Let me know if these changes close out this conversation.

ethan-moss:
I've raised #148 to capture a feature request picked up during the dev of this branch. It's for `defence.py`'s `check_list()` to be generalised to any iterable. This feature would be useful for further defence checks on iterable input arguments.

cbrowning04 (approved):
Great work finishing off the changes I had requested @ethan-moss! Everything is working well and I believe this branch is ready to be merged

PR #150, opened 2023-09-26
70 optional arg for viz stops to only plot stop ids in stop timestxt

### LINKED ISSUE #70: optional arg for `viz_stops()` to only plot `stop_id`'s in `stop_times.txt`

Add an optional argument to the [`viz_stops()`](https://github.com/datasciencecampus/transport-network-performance/blob/dev/src/transport_performance/gtfs/validation.py#L173), used when `geom="points"`, to only plot `stop_id`'s in `stop_times.txt`. 

This will correspond to the case where the stops are parts of trips within the GTFS data. Currently, it plots all stops in `stops.txt` regardless of whether or not they are references in `stop_times.txt`. This means stops with location_types not `nan` or `0` are also being visualised (e.g. stations, entrances/exits etc.). See the [GTFS spec for `stops.txt`](https://developers.google.com/transit/gtfs/reference#stopstxt) for more details.

Use optional argument to create a unique set of stops from `stops.txt` if and only if they are in `stop_times.txt`. This can be achieved by passing `feed.stops[feed.stops.stop_id.isin(feed.stop_times.stop_id.unique)].stop_id.to_list()` into the plotting function when the optional argument is given.

r-leyshon:
## Description

`GtfsInstance.viz_stops()` has an additional parameter `filtered_only=True` by default. This method will now write a HTML file based on the `stop_id`s appearing in the stop_times.txt only. Setting `filtered_only=False` returns the previous behaviour of plotting all `stop_id`s referenced within stops.txt.

Fixes #70 

## Motivation and Context

* After the feed is filtered using gtfs_kit, visualise only the points referenced within stop_times.
* Hull and point visualisations should agree.
* Refactoring has brought the `viz_stops` method up to date with defence utils.

## Type of change

## How Has This Been Tested?

Test configuration details:
* OS: macos
* Python version: 3.9.13

## Advice for reviewer

Advise running the following commands and checking the behaviour is consistent:
```
from transport_performance.gtfs.validation import GtfsInstance
out = "some_filenm.html"
feed = GtfsInstance()
feed.viz_stops(out)
# 1. View some_filenm.html to observe smaller subset of stops in a points map
feed.viz_stops(out, filtered_only=False)
# 2. View some_filenm.html to observe larger subset of stops in a points map
feed.viz_stops(out, geoms="hull")
# 3. View some_filenm.html to observe smaller subset of stops in a hull map
feed.viz_stops(out, geoms="hull")
# 4. View some_filenm.html to observe larger subset of stops in a hull map

"""Points 2 and 4 confirm that `GtfsInstance.feed.stops` table has not dropped rows where
`stop_id` does not appear within stop_times.txt"""



```

## Checklist:

### DIFF
--- a/src/transport_performance/gtfs/validation.py
+++ b/src/transport_performance/gtfs/validation.py
@@ -361,9 +361,8 @@ class GtfsInstance:
 
         # geoms defence
         geoms = geoms.lower().strip()
-        accept_vals = ["point", "hull"]
-        if geoms not in accept_vals:
-            raise ValueError("`geoms` must be either 'point' or 'hull.'")
+        ACCEPT_VALS = ["point", "hull"]
+        _check_item_in_list(geoms, ACCEPT_VALS, "geoms")
 
         try:
             m = self._produce_stops_map(
--- a/src/transport_performance/gtfs/validation.py
+++ b/src/transport_performance/gtfs/validation.py
@@ -111,6 +111,46 @@ def _create_map_title_text(gdf, units, geom_crs):
     return txt
 
 
+def _produce_stops_map(self, what_geoms, is_filtered, crs):
+    if what_geoms == "point":
+        if is_filtered:
+            plot_ids = self.feed.stop_times["stop_id"]
+        else:
+            plot_ids = self.feed.stops["stop_id"]
+        # viz stop locations
+        m = self.feed.map_stops(plot_ids)
+
+    elif what_geoms == "hull":
+        if is_filtered:
+            # filter the stops table to only those stop_ids present
+            # in stop_times, ensures hull viz agrees with point viz
+            stop_time_ids = set(self.feed.stop_times["stop_id"])
+            key_search = self.feed.stops["stop_id"].isin(stop_time_ids)
+            self.feed.stops = self.feed.stops[key_search]
+
+        # visualise feed, output to file with area est, based on stops
+        gtfs_hull = self.feed.compute_convex_hull()
+        gdf = gpd.GeoDataFrame(
+            {"geometry": gtfs_hull}, index=[0], crs="epsg:4326"
+        )
+        units = self.feed.dist_units
+        # prepare the map title
+        txt = _create_map_title_text(gdf, units, crs)
+        title_pre = "<h3 align='center' style='font-size:16px'><b>"
+        title_html = f"{title_pre}{txt}</b></h3>"
+        geo_j = gdf.to_json()
+        geo_j = folium.GeoJson(
+            data=geo_j, style_function=lambda x: {"fillColor": "red"}
+        )
+        m = folium.Map()
+        geo_j.add_to(m)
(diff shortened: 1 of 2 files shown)


src/transport_performance/gtfs/validation.py
```diff
@@ -242,9 +242,74 @@ def clean_feed(self):
...
+            if is_filtered:
+                # filter the stops table to only those stop_ids present
+                # in stop_times, this ensures hull viz agrees with point viz
+                stop_time_ids = set(self.feed.stop_times["stop_id"])
+                gtfs_hull = self.feed.compute_convex_hull(
+                    stop_ids=stop_time_ids
+                )
+            else:
+                # if not filtering, use gtfs_kit method
+                gtfs_hull = self.feed.compute_convex_hull()
```
CBROWN-ONS, line 283: This line is not covered by coverage

cbrowning04, line 283: This line is not covered by coverage

src/transport_performance/gtfs/validation.py
```diff
@@ -242,9 +242,74 @@ def clean_feed(self):
             # TODO: Issue 74 - Improve this to clean feed when KeyError raised
             print("KeyError. Feed was not cleaned.")
 
+    def _produce_stops_map(
```
cbrowning04, line 245: Not on this line but on line 379, we raise a `KeyError`. This may be more of a feature request but is it possible that we create our own stop_code column assuming that it only uses the stop_code column for data when you zoom into a singular stop.  
We could add any value to this columns such as `np.nan`

cbrowning04 (changes requested):
HI @r-leyshon. Very small changes requested, one of which is possible a feature request. Otherwise all good.  
I've also merged with dev and implemented a function from `defence.py` to save you having to do that.

src/transport_performance/gtfs/validation.py
r-leyshon, line 1: Nice one!

src/transport_performance/gtfs/validation.py
r-leyshon: good catch @CBROWN-ONS  - should now be patched.

src/transport_performance/gtfs/validation.py
r-leyshon: Hi @CBROWN-ONS - this likely deserves more of a conversation so that I can understand the scenario. But as it's not part of the diff in this PR, I'll put it in an issue with a 'needs triage' label.

cbrowning04: Sounds good. I'm happy to merge this PR now once the runners have finished 😄

src/transport_performance/gtfs/validation.py
cbrowning04: resolving as complete with commit f61b0c2

cbrowning04 (approved):

PR #190, opened 2023-10-23
186 filter to date method in gtfsinstance

### LINKED ISSUE #186: `filter_to_date()` method in `GtfsInstance`

Add a method to `GtfsInstance` that filters the GTFS data to a date of interest. This would dramatically reduce the size of GTFS file while not impacting the subsequent `r5py` analysis stage when using the GTFS data on the same date.

Method could be a shallow wrapper of existing `gtfs_kit` methods, also the suitability of this would need to be QAed/confirmed.

r-leyshon:
## Description

bbox_filter_gtfs optionally filters to a list of dates present within the gtfs calendar.

Fixes #186 

## Motivation and Context

Reducing GTFS archives to valid features (stops, shapes etc) on a given date appears to reduce the likelihood of encountering r5py errors when building transport network. It is currently thought that these errors are encountered due to specific fast travel issues, although further investigation and reproduction of the encountered error is ongoing.

## Type of change

bbox_filter_gtfs will continue to work as before, although specifying a populated list will run the `gtfs_kit.feed.restrict_to_date()` logic.

## How Has This Been Tested?

Defensive checks and test suite included. Also a test marked with runinteg to ensure the r5py trasport network can be built from the filtered GTFS file.

Test configuration details:
* OS: macos

## Advice for reviewer

Upon submission I noticed the test runners failing. After reviewing the fails, it became apparent that the current pytest invocation was incorrect and breaking. The changes to the CI workflows fix that and remove duplicated tests in all-os.tests.yml, where tests were being run twice:
1. pytest --runinteg
2. pytest

## Checklist:

- [ ] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/src/transport_performance/gtfs/gtfs_utils.py
+++ b/src/transport_performance/gtfs/gtfs_utils.py
@@ -20,13 +20,16 @@ from transport_performance.utils.defence import (
 from transport_performance.utils.constants import PKG_PATH
 
 
-def _validate_datestring(date_text, form="%Y%m%d"):
+def _validate_datestring(date_text: str, form: str = "%Y%m%d") -> None:
+    _type_defence(date_text, "date_text", str)
+    _type_defence(form, "form", str)
     try:
         datetime.strptime(date_text, form)
     except ValueError:
         raise ValueError(
             f"Incorrect date format, {date_text} should be {form}"
         )
+    return None
 
 
 def bbox_filter_gtfs(
--- a/src/transport_performance/gtfs/gtfs_utils.py
+++ b/src/transport_performance/gtfs/gtfs_utils.py
@@ -58,8 +58,8 @@ def bbox_filter_gtfs(
         What projection should the `bbox_list` be interpreted as. Defaults to
         "epsg:4326" for lat long.
     filter_dates: list, optional
-        A list of dates to restrict the feed to. Defaults to [] meaning that no
-        date filter will be applied.
+        A list of dates to restrict the feed to. Not providing filter_dates
+        means that date filtering will not be applied. Defaults to [].
 
     Returns
     -------
--- a/src/transport_performance/gtfs/gtfs_utils.py
+++ b/src/transport_performance/gtfs/gtfs_utils.py
@@ -44,7 +44,7 @@ def bbox_filter_gtfs(
     ],
     units: str = "km",
     crs: str = "epsg:4326",
-    filter_dates: Union[None, list] = None,
+    filter_dates: list = [],
 ) -> None:
     """Filter a GTFS feed to any routes intersecting with a bounding box.
 
@@ -66,8 +66,9 @@ def bbox_filter_gtfs(
     crs : str, optional
         What projection should the `bbox_list` be interpreted as. Defaults to
         "epsg:4326" for lat long.
-    filter_dates: Union[None, list], optional
-        A list of dates to restrict the feed to. Defaults to None.
+    filter_dates: list, optional
+        A list of dates to restrict the feed to. Defaults to [] meaning that no
+        date filter will be applied.
 
     Returns
     -------
@@ -93,7 +94,7 @@ def bbox_filter_gtfs(
         "crs": [crs, str],
         "out_pth": [out_pth, (str, pathlib.Path)],
         "in_pth": [in_pth, (str, pathlib.Path)],
-        "filter_dates": [filter_dates, (type(None), list)],
+        "filter_dates": [filter_dates, list],
     }
     for k, v in typing_dict.items():
         _type_defence(v[0], k, v[-1])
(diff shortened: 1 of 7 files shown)


src/transport_performance/gtfs/gtfs_utils.py
```diff
@@ -19,6 +20,15 @@
...
 
+def _validate_datestring(date_text, form="%Y%m%d"):
+    try:
+        datetime.strptime(date_text, form)
+    except ValueError:
+        raise ValueError(
+            f"Incorrect date format, {date_text} should be {form}"
+        )
+
+
```
cbrowning04: Is it possible to include type hinting + type defences here. A possible (optional) implementation might be:  
```python
def _validate_datestring(date_text: str, form: str="%Y%m%d") -> None:
    _type_defence(date_text, "date_text", str)
    _type_defence(form, "form", str)
    try:
        datetime.strptime(date_text, form)
    except ValueError:
        raise ValueError(
            f"Incorrect date format, {date_text} should be {form}"
        )
    return None
```

cbrowning04: After further consideration, I would also suggest adding some sort of docstring for this function. I appreciate that it is a private function, however even only a description of the function would be useful.

cbrowning04: Another comment on this function, sorry!  
I believe that this function could be useful in other modules, not just GTFS. I would support moving this function to a file such as defence.py.

src/transport_performance/gtfs/gtfs_utils.py
```diff
@@ -53,6 +66,9 @@ def bbox_filter_gtfs(
     crs : str, optional
         What projection should the `bbox_list` be interpreted as. Defaults to
         "epsg:4326" for lat long.
+    filter_dates: list, optional
+        A list of dates to restrict the feed to. Defaults to [] meaning that no
+        date filter will be applied.
```
cbrowning04: Most likely a 'won't change' but I'd like your opinion on this.  
Do you think we should alter this docstring to match other dosctrings in the form of having 'defaults to' as the last part of a param description (without the 'meaning that no
date filter will be applied.' afterwards)

src/transport_performance/gtfs/gtfs_utils.py
```diff
@@ -19,6 +20,15 @@
...
+def _validate_datestring(date_text, form="%Y%m%d"):
+    try:
+        datetime.strptime(date_text, form)
+    except ValueError:
+        raise ValueError(
+            f"Incorrect date format, {date_text} should be {form}"
+        )
+
+
 def bbox_filter_gtfs(
```
cbrowning04: Do we want to allow the user to pass a date format to this function? This would be helpful for the user and a good QOL update.

To allow for comparisons with the GTFS date format (%Y%m%d) we could convert to this format within the function if it isn't the passed format.

cbrowning04 (changes requested):
Hi @r-leyshon . Great work on this. Everything is working as expected.  
I have a few small comments however they aren't huge things to clear up.  

On the r5py stuff, from eyeballing it, it looks good to me. We might however want to get someone else to quickly run the test on their machine to ensure it works.

r-leyshon:
> Hi @r-leyshon . Great work on this. Everything is working as expected. I have a few small comments however they aren't huge things to clear up.
> 
> On the r5py stuff, from eyeballing it, it looks good to me. We might however want to get someone else to quickly run the test on their machine to ensure it works.

@SergioRec @ethan-moss can anyone help complete the review by running the tests locally on their machine?

src/transport_performance/gtfs/gtfs_utils.py
r-leyshon: This should be the current implementation - `_validate_datestring()` would be compatible with any valid C-format date code, whereas `bbox_filter_gtfs()` uses the internal with the required date format expected by `gtfs_kit`.

r-leyshon:
@CBROWN-ONS I have implemented all your suggestions except the final comment, which unless I have misunderstood is already the case. Many thanks for the good catches once again. Did you feel you needed a more in depth review of the tests or does a passing CI suite do enough to demonstrate the behaviour is as expected?

cbrowning04:
> @CBROWN-ONS I have implemented all your suggestions except the final comment, which unless I have misunderstood is already the case. Many thanks for the good catches once again. Did you feel you needed a more in depth review of the tests or does a passing CI suite do enough to demonstrate the behaviour is as expected?

Hi Rich. After reviewing your comments on the matter, I agree that we are fine not to accept a parameter for date format while filtering the GTFS.

Great work on the other changes. Regarding the r5py stuff, I'm happy to merge as the CI demonstrates that these tests pass and are operational. Let me know if you agree with this, and I will merge this into dev.

r-leyshon:
> > @CBROWN-ONS I have implemented all your suggestions except the final comment, which unless I have misunderstood is already the case. Many thanks for the good catches once again. Did you feel you needed a more in depth review of the tests or does a passing CI suite do enough to demonstrate the behaviour is as expected?
> 
> Hi Rich. After reviewing your comments on the matter, I agree that we are fine not to accept a parameter for date format while filtering the GTFS.
> 
> Great work on the other changes. Regarding the r5py stuff, I'm happy to merge as the CI demonstrates that these tests pass and are operational. Let me know if you agree with this, and I will merge this into dev.

Good to go from my end, thanks again.

cbrowning04 (approved):

PR #195, opened 2023-10-30
74 130 clean feed and route type cleaner

### LINKED ISSUE #72: develop `gtfs` cleaner for `Unrecognised column` warnings

`Unrecognised column` warnings are commonly raised by the `is_valid()` method, but are not resolved by cleaning. 

E.g. `Unrecognized column agency_noc` and `Unrecognized column vehicle_journey_code` are regularly raised when working with GTFS data extracted from DfT's BODS. These columns have nothing to do with the GTFS spec and can be safely removed from the feed (NOC is National Operator Code and journey_code is used by DfT to describe the bus type).

Add an extra optional argument to `clean_feed()` which accepts a lists of columns to drop. For each column provided, these can then be removed from and feed dataframe if the column is present. Giving the user the ability to define which ones can be removed safely is important here, since `gtfs_kit` doesn't handle the extended GTFS definitions and so dropping all unrecognised columns could lead to issues.

### LINKED ISSUE #73: develop `gtfs` cleaner to handle `repeated pair (trip_id, departure_time)` warnings

Calling `is_valid()` regularly leads to `repeated pair (trip_id, departure_time)` in `stop_times.txt` but these are not cleaned by `clean_feed()`. Further investigation shows that:
- typically this is because neighbouring rows in `stop_times.txt`, within a given trip, have duplicate trip_id, arrival_time, departure_time, and stop_id. It's clear this circumstance is an erroneous duplication.
- less often, there are neighbouring stops very close together, but they have different `stop_ids` - it's possible for these to be timetabled with the same time because of their proximity. Here is an example where a bus turns around and serves two stops on either side of the road (near Dingestow):
Screenshot 2023-08-14 at 18 35 05

In `stop_times.txt`, when records have duplicate trip_id, arrival_time, departure_time, and stop_id these can be safely dropped (represent the same trip, stop and time). Develop a clean which drops these from the stop_times feed, to prevent these warnings from remaining.

**Additional context**

### LINKED ISSUE #74: Improve `clean_feed()` when shape_id not available

`clean_feed()` is a shallow wrapper around `gtfs_kit`'s `clean()` function which implements 4 cleaning functions:
1. `clean_ids()`: strip whitespace from all string IDs and then replace every remaining whitespace chunk with an underscore
2. `clean_times()`: convert H:MM:SS time strings to HH:MM:SS time strings to make sorting by time work as expected.
3. `clean_route_short_names()`: In `feed.routes`, assign ‘n/a’ to missing route short names and strip whitespace from route short names. Then disambiguate each route short name that is duplicated by appending ‘-’ and its route ID. Note: this is the method that fixes the "Repeated pair (route_short_name, route_long_name)" warning. 
4. `drop_zombies()`: does the following:
    - Drop stops of location type 0 or NaN with no stop times.
    - Remove undefined parent stations from the parent_station column.
    - Drop trips with no stop times.
    - Drop shapes with no trips.
    - Drop routes with no trips.
    - Drop services with no trips.

`clean_feed()` and hence `clean()` will fail if there is no `shape_id` column in `trip.txt`. However, `drop_zombies()` is the only one that relies on that column, the 3 other cleaning functions work fine:
Screenshot 2023-08-15 at 10 57 39

Instead of returning without cleaning if `shape_id` is not present, this could be improved to instead only action the first 3 cleaning methods.

### LINKED ISSUE #130: invalid route type caveat

For route type == 200, can we include some text at the top, explaining that the message from gtfs_kit is not helpful, and that the route_type 200 is part of the proposed extended schema that they do not recognise? I think it’s worth doing though not super urgent as anyone using this is likely to encounter this route type in their GTFS.

some text at the top of the table when the issue is about route type 200 (coaches)

**Additional context**

image

### LINKED ISSUE #134: Check tables in each GTFS file and verify them
I can think of 2 other conditions we would want to check for raise:

1. Fewer files than expected
2. Additional files than expected (perhaps a warning is more appropriate here if we can test it without having it show up in test reports)

As logic not implemented, this can be a feature request. Would you agree?

_Originally posted by @r-leyshon in https://github.com/datasciencecampus/transport-network-performance/pull/112#discussion_r1325958601_

### LINKED ISSUE #180: fast travel cleaners tech-debt
Fast travel tech-debt raised during PR #125, associated with `cleaners.py`:

1.  Refactor `clean_multiple_stop_fast_travel_warnings()` and `clean_consecutive_stop_fast_travel_warnings()` into a common cleaning function:
    - Methods are similar, just the warning message and impacted tables are different.

2. `filter_gtfs_around_trip()` units optional argument should be "km" - or take from `gtfs.feed.dist_units` <--- probably best approach.

~~3. Use `pd.isna()` in `filter_gtfs_around_trip()` - handles strings, no need to try/except.~~

~~4. Add a unit test for `_gtfs_defence()` in the passing case (i.e. using a GTFS test fixture)~~

~~5. For `_add_validation_row()` add defence to ensure message is only either `{"warning", "error"}`~~

~~6. `drop_trip()` works with trip_ids that aren't in `stop_times`. This is OK since no cleaning is done, but it's slightly opaque for public function, maybe we can protect against this? Perhaps raise a warning to say not cleaning since no warnings/no valid trip_ids?~~

Points 3,4,5 & 6 resolved in #254 

7. Develop a unit test for the default speed bounds return statement (when the route_type is not a SPEED_BOUNDS key):
    - modify the GTFS fixture, perhaps replacing the route_type for a trip with a different value?

### LINKED ISSUE #181: gtfs `clean_feed()` pipeline

The sequence of GTFS cleaning steps is highly dependent on both ordering requirements and user needs. An ability to flexibly map GTFS cleaning steps to a user's request would be useful feature to give flexibility and control over cleaning methods.

A list of dicts could be used, where keys are the cleaning function and values are another dict corresponding to argument/value pairs for the cleaning functions, e.g.:

```python
gtfs_clean_pipeline = [
	{
		"clean_multiple_stop_fast_travel_warnings": {
			"max_distance": 10,
			...
		}
	},
	{
		"clean_consecutive_stop_fast_travel_warnings": {
			...
		}
	},
	...
]
```

**Additional context**

Feature request raising from PR #125

### LINKED ISSUE #182: close out fast travel TODO comments
The following TODO comments remain outstanding in `validators.py` and `cleaners.py`, associated with fast travel:
    - [ ]  Integrate fast travel results into HTML outputs (extended outputs)
    - [ ]  Add dtype defences to `_add_validation_row()`
    - [ ]  Add datatype defences in `filter_gtfs_around_trip()`

Raised during PR #125

cbrowning04:
## Description

This PR includes huge improvements to the cleaning/validating of a GTFS file.  

There are three main aspects of this PR:
1. A pipelined approach to cleaning and validating the GTFS
2. New cleaners and validators for `GtfsInstance`
3. GTFS validation/cleaning tech debt fixes

Fixes #130 
Fixes #74 
Fixes #181 
Fixes #134 
Fixes #72 
Fixes #182 
Fixes #180 
Fixes #73 

## Issue Distribution
After reviewing the issues fixed in this PR, I have decided to assign the issues to the following people. My main approach for this is limiting @ethan-moss to only the pipelined approach to cleaners/validators, along with tech debt issues raised by himself. The rest of the issues with be assigned to @r-leyshon . This list is subject to change however.  

Ethan
- #74 (intertwined with pipelined approach)
- #181
- #182
- #180

Rich  
- #130
- #134
- #72 
- #73

I also advise that Rich pick up lines that my not directly relate to the issue descriptions, such as additional utilities.

## File Distribution

Similar to above, the files/functions/classes must be allocated to certain reviewers.  

Ethan
- gtfs_utils.py (only function `_function_pipeline` and some fixed tech debt that is mentioned in the issues)
- validation.py, some tech debt resolved, `is_valid` and `clean_feed` pipeline approach

Rich
- cleaners.py
- gtfs_utils.py (all but the function `_function_pipeline` and tech debt)
- validation.py, changes to `html_report` and `_extended_validation`
- validators.py
- constants.py
- test_defence.py - added unit test requested in tech debt ticket

## Type of change

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected) (MAYBE?)

## How Has This Been Tested?

Test configuration details:
* OS: Windows 10
* Python version: 3.9.13
* Java version: N/A

## Checklist:


src/transport_performance/gtfs/gtfs_utils.py
```diff
@@ -336,3 +344,143 @@ def convert_pandas_to_plotly(
...
+    gtfs.validity_df = gtfs.validity_df.loc[
+        list(set(gtfs.validity_df.index) - set(index))
+    ]
+    return None
+
+
+def _function_pipeline(
+    gtfs, func_map: dict, operations: Union[dict, type(None)]
+) -> None:
+    """Iterate through and act on a functional pipeline."""
```
SergioRec, line 586: Missing docstring.

src/transport_performance/gtfs/validation.py
```diff
@@ -16,18 +16,15 @@
...
+import transport_performance.gtfs.cleaners as cleaners
+import transport_performance.gtfs.validators as gtfs_validators
+
 from transport_performance.gtfs.routes import (
     scrape_route_type_lookup,
     get_saved_route_type_lookup,
 )
+
+from transport_performance.gtfs.gtfs_utils import _function_pipeline
 from transport_performance.utils.defence import (
```
SergioRec, line 29: Would it be easier to just import defence with an alias instead of this many individual functions?

src/transport_performance/gtfs/validation.py
```diff
@@ -329,25 +370,28 @@ def get_gtfs_files(self) -> list:
...
-        self.validity_df = self.feed.validate()
-        if far_stops:
-            validate_travel_between_consecutive_stops(self)
-            validate_travel_over_multiple_stops(self)
+        _type_defence(validators, "validators", (dict, type(None)))
+        # create validity df
+        self.validity_df = pd.DataFrame(
+            columns=["type", "message", "table", "rows"]
+        )
+        _function_pipeline(
```
SergioRec, line 435: See https://docs.python.org/3/faq/programming.html#how-do-i-use-strings-to-call-functions-methods

You have chosen the first ("best") option, a dictionary assigning names to function so you can call those functions by name later. The issue I see with this in this instance is that `_function_pipeline` needs two arguments, first the hardcoded dict (which will need to be updated when functions change) and then another dict with the function name and arguments. When calling a function with no arguments, then `None` has to be used.

A potential way could be to use the second option in the link. This way, a toml file with the function names and arguments can be loaded into the script. There would be no need of hardcoded dict. Users would only need to change or add other toml files to get different behaviours. The function can be called like, e.g.:
```
getattr(module, "function_name")(**kwargs)
```

tests/gtfs/test_gtfs_utils.py
```diff
@@ -163,7 +170,7 @@ def test__add_validation_row_defence(self):
     def test__add_validation_row_on_pass(self):
         """General tests for _add_test_validation_row()."""
         gtfs = GtfsInstance(gtfs_pth=GTFS_FIX_PTH)
-        gtfs.is_valid(far_stops=False)
+        gtfs.is_valid(validators={"core_validation": None})
```
SergioRec, line 303: This kind of behaviour is not documented. According to `is_valid` docstrings, `validators` is "a dictionary of function name to kwargs mappings". However, in this case this is a dictionary of function name and `None`, which is not a keyword argument and is not a valid argument for `core_validation`.

I can see some logic in `_function_pipeline`, but I would not anticipate this behaviour without checking the source code.

If I understand correctly, if `None` is passed, then no keywords are passed into the function apart from `gtfs`. This is used as a way to select validation methods. So instead of selecting validation functions by changing the `func_map` argument in `_function_pipeline`, only functions passed to `validators` are run. Because none of the validators have arguments other than `gtfs`, then `None` must be passed, and some logic in `_function_pipeline` handles that behaviour. If my understanding is correct, this is a very convoluted way of selecting which validations to run. This would need to be clearly documented, and is related to my other comment about `func_map`.

src/transport_performance/gtfs/gtfs_utils.py
```diff
@@ -155,14 +160,20 @@ def _add_validation_row(
         An error is raised if the validity df does not exist
 
     """
-    # TODO: add dtype defences from defence.py once gtfs-html-new is merged
-    if "validity_df" not in gtfs.__dict__.keys():
-        raise AttributeError(
+    _gtfs_defence(gtfs, "gtfs")
+    _type_defence(_type, "_type", str)
+    _type_defence(message, "message", str)
```
SergioRec, line 292: Missing defence for `table`

src/transport_performance/gtfs/cleaners.py
```diff
@@ -61,44 +80,58 @@ def drop_trips(gtfs, trip_id: Union[str, list, np.ndarray]) -> None:
     return None
 
 
-def clean_consecutive_stop_fast_travel_warnings(
-    gtfs, validate: bool = False
-) -> None:
-    """Clean 'Fast Travel Between Consecutive Stops' warnings from validity_df.
+def _clean_fast_travel_preperation(gtfs, warning_re: str) -> pd.DataFrame:
```
SergioRec: Typo: should be "preparation". Not changing this as it'll probably require other changes in the codebase where this function is used.

src/transport_performance/gtfs/cleaners.py
```diff
@@ -61,44 +80,58 @@ def drop_trips(gtfs, trip_id: Union[str, list, np.ndarray]) -> None:
...
+        ),
+    )
+    needed_warning = _get_validation_warnings(gtfs, warning_re)
+    return needed_warning
+
 
-    if validate:
-        gtfs.is_valid()
+def clean_consecutive_stop_fast_travel_warnings(gtfs) -> None:
+    """Clean 'Fast Travel Between Consecutive Stops' warnings from validity_df.
```
SergioRec, line 120: When running this function in the Wales GTFS, it seems that it also purges instances of fast travel between multiple stops.

So when running:
```
gtfs.is_valid()
clean_consecutive_stop_fast_travel_warnings(gtfs)
gtfs.is_valid()
```
It gets rid of all `Fast Travel Between Consecutive Stops` warnings as well as all `Fast Travel Over Multiple Stops` warnings. This causes that running `clean_multiple_stop_fast_travel_warnings` after `clean_consecutive_stop_fast_travel_warnings` has no effect. I'm not sure if there may be instances where this is not the case, but I've tried it with GTFS from Wales, Scotland and London.

Considering this, it is unlikely that we'll run `clean_consecutive_stop_fast_travel_warnings` in isolation, as it could leave some problematic trips undetected. Would it make sense to merge them into a single function and apply them sequentially? Or run `clean_multiple_stop_fast_travel_warnings` by default, and have an optional flag to also run the other step if requested, always after the former? This way we would be refactoring these two functions into a single one and ensuring cleaning steps are applied in the right order.

src/transport_performance/gtfs/gtfs_utils.py
```diff
@@ -182,7 +193,6 @@ def filter_gtfs_around_trip(
     gtfs,
     trip_id: str,
     buffer_dist: int = 10000,
```
SergioRec, line 322: If this function is not using meters anymore, this default needs to be changed or removed. Currently it's using 10,000 km.

src/transport_performance/gtfs/gtfs_utils.py
```diff
@@ -182,7 +193,6 @@ def filter_gtfs_around_trip(
     gtfs,
```
SergioRec, line 320: This function is not being used anywhere in cleaners or validators. What is its purpose?

tests/utils/test_defence.py
```diff
@@ -244,13 +246,18 @@ def test_check_parents_dir_exists(self, tmp_path):
...
     with pytest.raises(
         TypeError,
         match=re.escape(
             "'test' expected a GtfsInstance object. " "Got <class 'str'>"
         ),
     ):
         _gtfs_defence("tester", "test")
+    # passing test
+    with does_not_raise():
+        gtfs = GtfsInstance(here("tests/data/chester-20230816-small_gtfs.zip"))
```
SergioRec, line 259: Why use a new fixture with data from Chester instead of the existing Newport fixture? I'm sure there's a good reason, just asking for clarification!

src/transport_performance/gtfs/validators.py
```diff
@@ -26,6 +31,28 @@
     200: 120,
```
SergioRec, line 31: Probably worth adding a warning when route type not found and using the 200 km/h default?

SergioRec (commented):
Review started but not finished. Since we're delaying this PR to complete others, just pushing the comments I have so far in case I don't come back to this PR in the future.

tests/utils/test_defence.py
cbrowning04: Chester fixture is needed for fast travel warnings.

r-leyshon:
Migrated to https://github.com/datasciencecampus/assess_gtfs/pull/20

PR #200, opened 2023-11-09
173 analyse network

### LINKED ISSUE #173: develop initial `analyse_network` module

Develop an `analyse_network` module that takes the relevant sections from `e2e.py`. 

This could look like an `r5py` wrapper module that can batch inputs/outputs to handle larger areas. It could also including a plotting method to be able to visualise and QA the results more readily.

**Additional context**

### LINKED ISSUE #211: Refactoring of `od_matrix` in `analyse_network.py`.

Currently, function `od_matrix` does distance filtering even when using all origins. This has no effect in the actual o-d matrix output (as r5py creates all combinations again), and adds a few minutes of processing time that could be saved.

Add a flag to use either one origin per batch, or all. If one, use the loop to go through origins one by one doing distance filtering. If all, run the transport performance with all data.

Unit tests will need to be adapted accordingly.

**Additional context**

This was raised by @ethan-moss during PR review. We agreed the PR would be paused while this was implemented.

SergioRec:
## Description
Module analyse_network.

Closes #173 
Fixes #211

## Motivation and Context
This module is a wrapper for r5py to calculate an origin-destination matrix to feed into the matrix module. It takes the outputs from the urban_centres and rasterpop modules.

Module allows selection of a different number of origins to calculate O-D matrices in batches. It saves the outputs as parquet files, one per origin batch. However, it will split any parquet file exceeding a certain size.  

## Type of change

## How Has This Been Tested?
Module run locally, outputs checked manually.
Unit tests run locally.

Test configuration details:
* OS: MacOS Ventura
* Python version: 3.9.13
* Java version: 11
* Python management system: pip/conda

## Advice for reviewer

Run times:
- Newport: all origins in batch, ~3 minutes
- Marseille: all origins in batch, ~10 minutes
- Leeds: all origins in batch, ~ 1 hour
- London: one origins per batch, ~40 hours

## Checklist:

## Additional comments

I've chosen to not explicitly ask for keyword arguments from r5py and leave them as `**kwargs`. This may not be a good choice, but I wasn't sure if arbitrarily including some arguments and not others would be messy. Happy to change it.

Regarding the defence against dates not being in the GTFS (which also triggers when no valid GTFS is passed and `TransportMode.TRANSIT` is included), it's not straightforward to access any attribute within `TransportNetwork` or `TravelTimeMatrixComputer` that exposes GTFS information. The current solution catches the specific runtime warning and raises an `IndexError` when present.


notebooks/urban_centres/buffer_experiments.py
```diff
@@ -0,0 +1,253 @@
+"""Notebook to check buffer and distance to edge cells."""
```
ethan-moss, line 1: What do you think about committing this file to `dev`? I'm happy to, but wondering if this is a duplicate of `notebooks/urban_centres/explore_uc_buffer.py`?

src/transport_performance/analyse_network.py
```diff
@@ -0,0 +1,441 @@
...
+        d._check_iterable(
+            iterable=gtfs,
+            param_nm="gtfs",
+            iterable_type=list,
+            check_elements=True,
+            exp_type=(str, pathlib.Path),
+        )
+        for path in gtfs:
+            d._is_expected_filetype(path, "gtfs", exp_ext=".zip")
+
```
ethan-moss: As part of the `__init__` defence check, what do you think about ensuring the CRS of `gdf` is `EPSG:4326`? It could cause a lot of problems for this module if that isn't the case (`r5py` doesn't work with it and the Haversine distance measurements will be incorrect, for example).

requirements.txt
```diff
@@ -31,6 +31,8 @@ kaleido
 numpy>=1.25.0 # test suite will fail if user installed lower than this
 sphinx
 sphinx-rtd-theme
+haversine
+dask
```
ethan-moss, line 34: `pyarrow` < 14.0.1 has a security vulnerability that's fixed in >=14.0.1 (https://security.snyk.io/package/pip/pyarrow). I got a warning when running an older version, and updating doesn't cause any issues with the codebase. OK to fix the version to >=14.0.1 here?

src/transport_performance/analyse_network.py
```diff
@@ -0,0 +1,441 @@
...
+        d._type_defence(od_matrix, "od_matrix", pd.DataFrame)
+        d._type_defence(out_name_func, "out_name_func", str)
+        d._type_defence(npartitions, "npartitions", int)
+        d._check_parent_dir_exists(out_path, "out_path", create=True)
+
+        ddf = dd.from_pandas(od_matrix, npartitions=npartitions)
+
+        filename = f"batch-{out_name_func}"
+        filefunc = lambda x: filename + f"-{x}.parquet"  # noqa: E731
+        ddf.to_parquet(out_path, name_function=filefunc)
```
ethan-moss, line 478: When `dask` writes to file, it modifies the index on re-reading: 
Screenshot 2023-12-06 at 16 09 45

What do you think about [renaming the index](https://docs.dask.org/en/latest/generated/dask.dataframe.Index.rename.html)/removing the name before writing?

src/transport_performance/analyse_network.py
```diff
@@ -0,0 +1,441 @@
...
+                )
+
+                self._save_to_parquet(
+                    od_matrix, str(min(sel_orig)), out_path, partitions
+                )
+        else:
+            origin_gdf = self.gdf.copy()
+            dest_gdf = self.gdf[
+                self.gdf[destination_col] == True  # noqa: E712
+            ].copy()
```
ethan-moss: This could be changed to:
```python
dest_gdf = self.gdf[self.gdf[destination_col]].copy()
```
Then there shouldn't be a need for the noqa comment. The condition inside the [] is equivalent to '== True'.

src/transport_performance/analyse_network.py
```diff
@@ -0,0 +1,441 @@
...
+        -------
+        int
+            Required partitions for the parquet file.
+
+        """
+        # defences
+        d._type_defence(df, "df", pd.DataFrame)
+        d._type_defence(partition_size, "partition_size", int)
+
+        mem_usage = sum(df.memory_usage(deep=True, index=False))
```
ethan-moss, line 441: What is the reason for setting `index=False` here? My thinking is we should include the index when calculating the total memory usage, since physically is contributes to the dataframe?

src/transport_performance/analyse_network.py
```diff
@@ -0,0 +1,441 @@
...
+                    "a valid GTFS."
+                )
+
+        return travel_time_matrix_computer.compute_travel_times()
+
+    def _gdf_batch_origins(
+        self,
+        gdf: gpd.GeoDataFrame,
+        destination_col: str,
+        distance: float,
```
ethan-moss: Perhaps `distance` could be of type `float` or `int`? That way, entering say `10` wouldn't trigger the type error?

src/transport_performance/analyse_network.py
```diff
@@ -0,0 +1,441 @@
...
+        # defences
+        d._type_defence(df, "df", pd.DataFrame)
+        d._type_defence(partition_size, "partition_size", int)
+
+        mem_usage = sum(df.memory_usage(deep=True, index=False))
+        est_parts = ceil(mem_usage / (partition_size * 1e6))
+
+        return est_parts
+
+    def _save_to_parquet(
```
ethan-moss, line 446: This overwrites any existing data in a folder (should it exist already). Do you think it could be beneficial to raise a warning here if the directory already has some parquet files? Maybe a note in the docstring would suffice? Maybe we do nothing?

src/transport_performance/analyse_network.py
```diff
@@ -0,0 +1,441 @@
...
+                f"got {num_origins}"
+            )
+
+        # get sources and destinations
+        # define destinations when `destination_col` is true
+        orig_gdf = gdf.copy()
+        geometry_col = gdf.geometry.name
+        # TODO: add option to include all rows as destinations?
+        dest_gdf = (
+            gdf[gdf[destination_col] == True]  # noqa: E712
```
ethan-moss: As above, the `==True` part isn't needed here, then the `noqa` comment could go to.

src/transport_performance/analyse_network.py
```diff
@@ -0,0 +1,441 @@
...
+                f"`num_origins` should be between 1 and {len(gdf)}, "
+                f"got {num_origins}"
+            )
+
+        # get sources and destinations
+        # define destinations when `destination_col` is true
+        orig_gdf = gdf.copy()
+        geometry_col = gdf.geometry.name
+        # TODO: add option to include all rows as destinations?
+        dest_gdf = (
```
ethan-moss: At this stage, it's possible for users to provide an invalid column and this filtering stage wouldn't throw an error. For example `destination_col` could be incorrectly set to `"id"`, and the filtering condition would filter only to the case where the `"id"` columns is one:
Screenshot 2023-12-06 at 16 21 23
This is only possible though, since `destination_col` can be set by users in the `od_matrix()` call. We could maybe add a `dtype` check to the request column to make sure it's `bool`?

src/transport_performance/analyse_network.py
```diff
@@ -0,0 +1,441 @@
...
+            gdf[gdf[destination_col] == True]  # noqa: E712
+            .reset_index()
```
ethan-moss: The index is being reset here OK, but it's not dropped - an 'index' column still remains in the datafame. As far as I can tell, this isn't used in subsequent code, could be perhaps use the `drop=True` argument here to save on an unnecessary column?

src/transport_performance/analyse_network.py
```diff
@@ -0,0 +1,441 @@
...
+            full_gdf["distance"] = self._haversine_df(
+                full_gdf,
+                f"{geometry_col}_orig",
+                f"{geometry_col}_dest",
+                unit=unit,
+            )
+
+            # filters out pairs where distance is over threshold
+            sel_dest = list(
+                full_gdf[full_gdf["distance"] < distance]["id_dest"].unique()
```
ethan-moss: The equivalent condition in `metrics` uses `<=`. We should update either the `metrics` module or this line to match. My feeling would be to use `<=` so everything greater than this is excluded, so a change here would suit that. I can't think of any major pros/cons for either way to be honest, since this really is a very very rare corner case for the distance to be exactly 11.25Km, say). What do you think?

ethan-moss (commented):
I've completed a review of the source code updates. The main functionalities all LGTM. I've tested them with a few urban centres and:
- we're getting consistent results to those we got previously, and
- transport performance metrics are identical whether calculated using the batched mode or not.

Just sharing initial comments at this stage, as it would be useful to pick out which of these (if any at all) need to be implemented right away, tagged as tech-debt, or treat as 'won't fix'. Let me know your thoughts on this.

In the meantime I'll continue to review the unit tests and add comments

src/transport_performance/analyse_network.py
SergioRec: Doing nothing may lead to some issues. If the number of files being produced is less than the number of files already in the folder (e.g. we're running all origins when we run single origins before), it will only overwrite some of the files leaving the others. This will cause duplication when loading the parquet files for the metrics.

Not sure what would be the best way to go. I think we should definitely empty the folder before saving new files, but not sure how the user should be made aware of this. Should we raise an error if the folder has content? Should we add a flag to empty the folder? Happy to discuss.

src/transport_performance/analyse_network.py
SergioRec: Added defence. If crs is not epsg: 4326, a warning is raised and gdf is reprojected.

src/transport_performance/analyse_network.py
SergioRec: Done.

src/transport_performance/analyse_network.py
SergioRec: Done.

src/transport_performance/analyse_network.py
SergioRec: Done.

src/transport_performance/analyse_network.py
SergioRec: Done.

requirements.txt
SergioRec: Done.

src/transport_performance/analyse_network.py
SergioRec: Added check for `bool dtype`. Added tests.

src/transport_performance/analyse_network.py
SergioRec: Done.

notebooks/urban_centres/buffer_experiments.py
ethan-moss: Discussed and will commit with a slight different file name to make it more explict.

src/transport_performance/analyse_network.py
ethan-moss: We can treat this as an enhancement, potentially a utils function that reads in OD matrix results and handles the index at the same time. Feature request to be opened.

src/transport_performance/analyse_network.py
ethan-moss: Discussed and can proceed in 2 steps:
1. Check if provided dir has content (any parquet files) within the __init__ and raise.
2. Feature request to handle existing files without raising (e.g., move to a hold area, rename dir etc.).

SergioRec: I have added the defence in the `__init__`, which raises a `NotImplementedError` if files are in the provided output directory.

Tests have become a bit more complicated as a result. To create the `dummy_transport_network` fixture we need to create a module-scoped temp path. However, to test i/o behaviour we need to use a function-scoped temp path.

src/transport_performance/analyse_network.py
ethan-moss: Change has been made and is performing as expected - unit tests are passing + the OD matrix and transport performance results are identical to those obtained previously. Will close this conversation.

src/transport_performance/analyse_network.py
ethan-moss: Change has been made and is performing as expected - unit tests are passing + the OD matrix and transport performance results are identical to those obtained previously. Will close this conversation.

src/transport_performance/analyse_network.py
ethan-moss: Change has been made and is performing as expected - unit tests are passing + the OD matrix and transport performance results are identical to those obtained previously. Will close this conversation.

src/transport_performance/analyse_network.py
ethan-moss: Change has been made and is performing as expected - unit tests are passing + the OD matrix and transport performance results are identical to those obtained previously. Will close this conversation.

src/transport_performance/analyse_network.py
ethan-moss: This LGTM. It raises the expected error when an incorrect type is given, and setting the distance to the same value gives identical OD matrix results for the int and float case. Happy to close this conversation.

src/transport_performance/analyse_network.py
ethan-moss: Added commit `5a7b448` to move these checks to the start of `od_matrix()` so that the defence checks are done for both batched and non-batched modes. Unit tests updated to reflect this change too. Tested locally, and expected errors are being raised as needed. Happy to close this conversation. Thanks again for your help!

src/transport_performance/analyse_network.py
ethan-moss: Ticket #221 raised to capture this feature request. Closing the conversation.

src/transport_performance/analyse_network.py
ethan-moss: Step 1. above has been implemented and is working exactly as expected - raises an Error when parquet files already exist in the output folder to prevent overwriting. This can be circumnavigated by calling `od_matrix` with different configurations after an `AnalyseNetwork` instance has been instantiated. However, this approach does not align with the design choice made to priorities error raising as soon as possible. Nor is it, at this time, the expected way to use this class. For these reasons, plus the fact the class if functioning as needed, this last point can be considered as tech-debt and resolved at a later date. Ticket #222 has been raised to capture this tech-debt. Closing the conversation with this ticket.

src/transport_performance/analyse_network.py
ethan-moss: This is working as expected - warns when a `gdf` with an incorrect CRS is provided and the `geometry` is reprojected to EPSG:4326.

Back to back testing with an EPSG:4326 input and the same input re-projected to EPSG:27700 input conducted, and it seems re-projection introduces very very slight differences in the centroid coordinates. This isn't significant (down in the very low decimal places), nor is it controllable - it's an artifact of the re-projection calculations themselves. It's also likely compounded by this back-to-back test containing an input which was re-projected twice.

Closing this conversation based on these points.

ethan-moss (approved):
I've reviewed this PR and all LTGM - thanks very much for putting this together, it's a really great job!

As well as reviewing the new features and tests in detail, i've also run back-to-back comparisons with previous results (when this module wasn't available). This gave identical outputs for multiple urban centres. This was also identical in terms of performance for the non-batched approach. I also compared transport performance outputs for batched and non-batched methods - again identical results for multiple urban centres.

Only 2 pieces of tech-debt remain (they do not impact the core functioning of `analyse_network` and will be dealt with as smaller tickets):
1. Possible to overwrite parquet outputs when running `od_matrix()` with different configurations - #222
2. When batching, if there are no destination centroids within distance of the origin then it writes empty parquet files - #225

Also one enhancement off the back of this PR:
1. OD matrix index naming when writing to parquet - #221

Happy to merge to `dev`. Thanks very much again for your help with this.

PR #207, opened 2023-11-17
Multi gtfs instance

### LINKED ISSUE #204: rename bbox_filter_gtfs

The function now offers date filtering. Better to make consistent with the OSM module `filter_osm` function.

**Additional context**

cbrowning04:
## Description

This PR aims to allow for cleaning, validation, filtering, saving and visualising multiple GTFS files at once. It also refactors `bbox_filter_gtfs`. This function has remained however and is now depreciated.

Fixes #204 

## Motivation and Context

This motivation of this PR is to validate and clean multiple GTFS files in order to complete end to end pipeline runs.

## Type of change

## How Has This Been Tested?

Test configuration details:
* OS: Windows 10
* Python version: 3.9.13
* Java version: N/A

## Checklist:


src/transport_performance/gtfs/gtfs_utils.py
```diff
@@ -7,20 +7,135 @@
...
     _check_iterable,
     _type_defence,
     _validate_datestring,
     _enforce_file_extension,
+    _gtfs_defence,
 )
 from transport_performance.utils.constants import PKG_PATH
 
 
+def _validate_bbox(bbox: list) -> None:
```
r-leyshon: `_validate_bbox()` -> `_validate_bbox_list()` I'm guessing the other devs who tend to work with polygons may get distracted by the name.

src/transport_performance/gtfs/gtfs_utils.py
```diff
@@ -7,20 +7,135 @@
...
+    """Small function to ensure bbox coords are in the correct order."""
+    _type_defence(bbox, "bbox", list)
```
r-leyshon, line 31: doesn't look like we need this - this only gets executed in `filter_gtfs()` if bbox is a list.

src/transport_performance/gtfs/gtfs_utils.py
```diff
@@ -7,20 +7,135 @@
...
+        raise ValueError(
+            f"BBOX ymin ({bbox[1]}) is greater than ymax ({bbox[3]})"
+        )
+    return None
+
+
+def filter_gtfs(
+    gtfs: "GtfsInstance",
+    bbox: Union[GeoDataFrame, list, None] = None,
+    crs: str = "epsg:4326",
```
r-leyshon: crs can also work with an integer. "epsg:4326" == 4326

src/transport_performance/gtfs/gtfs_utils.py
```diff
@@ -42,6 +157,8 @@ def bbox_filter_gtfs(
 
     Optionally filter to a list of given dates.
 
+    WARNING: THIS FUNCTION IS DEPRECATED AND WILL INEVITABLY BE REMOVED.
```
r-leyshon, line 166: Not sure we need to deprecate before an official release. Perhaps other team members have requested this. In which case, we should raise a deprecation warning.

src/transport_performance/gtfs/gtfs_utils.py
cbrowning04: To give some insight on this, this was added to avoid a complexity hook.

r-leyshon: sorry, I mean specifically the _type_defence line.

cbrowning04: Oh right. Happy to remove this or allow it to remain as it might be useful in the future if we implement this function again. Either way is okay wth me.

src/transport_performance/gtfs/gtfs_utils.py
cbrowning04: Good shout.

src/transport_performance/gtfs/validation.py
```diff
@@ -212,6 +216,12 @@ class GtfsInstance:
         Returns the `daily_route_summary` attribute.
     html_report()
         Generate a HTML report describing the GTFS data.
+    save()
+        Save the current GtfsInstance().
+    filter_to_date()
+        Filter a GtfsInstance to a specific dates or set of dates.
+    filter_to_bbox()
+        Crop a GtfsInstance to a given bbox.
```
r-leyshon, line 224: Worth discussing why GtfsInstance would need to separate out the bbox / date filtering logic as separate methods, unsure if I'm missing something about how this needs to operate.

src/transport_performance/gtfs/validation.py
```diff
@@ -1637,3 +1647,74 @@ def html_report(
...
+    ) -> None:
+        """Very shallow wrapper around filter_gfts().
+
+        Filters GTFS to a bbox.
+
+        Parameters
+        ----------
+        bbox : Union[list, GeoDataFrame]
+            The bbox to filter the GTFS to. Leave as none if the GTFS does not
+            need to be cropped. Format - [xmin, ymin, xmax, ymax]
```
r-leyshon, line 1708: With a `filter_to_bbox` method, should we raise instead if no bbox is provided?

src/transport_performance/gtfs/multi_validation.py
```diff
@@ -0,0 +1,476 @@
...
+        Create a summary of all of the routes throughout all GTFS files.
+    summarise_routes()
+        Create a summary of all of the trips throughout all GTFS files.
+
+    Raises
+    ------
+    TypeError
+        'path' is not of type string or list.
+    FileCountError
+        The glob string used for glob.glob does not find at least 2 files.
```
r-leyshon: This indicates that the new class won't work for single GTFS files, right? Could be that this class becomes the way to handle GTFS with detection of the number of GTFS files found controlling the logic used (eg 1 GTFS file use single GTFS, more than one use multi GTFS.)

src/transport_performance/gtfs/multi_validation.py
```diff
@@ -0,0 +1,476 @@
...
+        _type_defence(path, "path", (str, list))
+        # defend a glob string
+        if isinstance(path, str):
+            gtfs_paths = glob.glob(path)
+            if len(gtfs_paths) < 2:
+                raise FileCountError(
+                    f"At least 2 files expected at {path}. Found "
+                    f"{len(gtfs_paths)}"
+                )
+            path = gtfs_paths
```
r-leyshon, line 104: is this needed?

src/transport_performance/gtfs/multi_validation.py
```diff
@@ -0,0 +1,476 @@
...
+        # check all paths are zip files
+        for i, pth in enumerate(path):
+            _is_expected_filetype(pth, f"path[{i}]", True, ".zip")
+
+        self.paths = path
+        # instantiate the GtfsInstance's
+        self.instances = [GtfsInstance(fpath) for fpath in path]
```
r-leyshon, line 113: That's a neat solution. I'm wondering if a clearer naming of GtfsInstance would now be SingleGtfs and MultiGtfs?

src/transport_performance/gtfs/multi_validation.py
```diff
@@ -0,0 +1,476 @@
...
+
+    def save(self, dir: Union[pathlib.Path, str]) -> None:
```
r-leyshon: save_feeds?

src/transport_performance/gtfs/multi_validation.py
```diff
@@ -0,0 +1,476 @@
...
+            )
+            for p in self.paths
+        ]
+        progress = tqdm(zip(save_paths, self.instances), total=len(self.paths))
+        for path, inst in progress:
+            progress.set_description(f"Saving at {path}")
+            inst.save(path)
+        return None
+
+    def clean_feed(self, clean_kwargs: Union[dict, None] = None) -> None:
```
r-leyshon: clean_feeds?

src/transport_performance/gtfs/multi_validation.py
```diff
@@ -0,0 +1,476 @@
...
+        # defences
+        _type_defence(dates, "dates", (str, list))
+        # convert to normalsed format
+        if isinstance(dates, str):
+            dates = [dates]
+        # filter gtfs
+        progress = tqdm(zip(self.paths, self.instances), total=len(self.paths))
+        for path, inst in progress:
+            progress.set_description(f"Filtering GTFS from path {path}")
+            inst.filter_to_date(dates=dates)
```
r-leyshon: I'm wondering if we should try/except cases where at least one of the GTFS doesn't contain the correct dates? the except case would raise a waring but continue to clean the rest of the feeds. Alternatively raise an exception - perhaps discuss with the rest of the team?

src/transport_performance/gtfs/multi_validation.py
r-leyshon: Checking in with dev team this morning, this class needs to work with single or multiple GTFS instance without adjusting the code, so a little light refactoring to ensure this is the case would be required.

src/transport_performance/gtfs/multi_validation.py
r-leyshon: Opinion from dev team this morning is that they need to raise if at least one of the feeds has the wrong dates. the filter_to_dates method has implemented a raise for that condition, however a more helpful warning indicating the precise filename of the bad GTFS would be required.

src/transport_performance/gtfs/multi_validation.py
```diff
@@ -0,0 +1,476 @@
...
+        return None
+
+    def filter_to_bbox(
```
r-leyshon, line 292: Feature request this morning from the team was to warn if one of the GTFS files contains no GTFS following the filtering. We discussed what that meant in practice. Check the GTFS tables and ensure that none of them have zero rows.

src/transport_performance/gtfs/gtfs_utils.py
```diff
@@ -7,20 +7,135 @@
...
+    ------
+    ValueError
+        Raised if any of the dates passed to 'filter_dates' isn't present in
+        the gtfs
+
+    """
+    # defences
+    _gtfs_defence(gtfs, "gtfs")
+    _type_defence(bbox, "bbox", (GeoDataFrame, list, type(None)))
+    _type_defence(crs, "crs", str)
```
r-leyshon: needs to accept integer also

r-leyshon (changes requested):
Thanks @CBROWN-ONS, a neat implementation. Couple of minor things and behaviour specified by rest of the team.

src/transport_performance/gtfs/gtfs_utils.py
cbrowning04: Going to resolve this. Leaving in type defence for extra measures. Un-noticeable computational expenses and also protects for any future uses.

src/transport_performance/gtfs/multi_validation.py
cbrowning04: Done.

src/transport_performance/gtfs/multi_validation.py
cbrowning04: Done.

src/transport_performance/gtfs/gtfs_utils.py
cbrowning04: Done

src/transport_performance/gtfs/multi_validation.py
cbrowning04: Done

src/transport_performance/gtfs/gtfs_utils.py
cbrowning04: Warning added

src/transport_performance/gtfs/gtfs_utils.py
cbrowning04: Done

src/transport_performance/gtfs/multi_validation.py
cbrowning04: done

src/transport_performance/gtfs/multi_validation.py
cbrowning04: Done

r-leyshon: Hi @CBROWN-ONS  - I can't see where this is implemented. Quick scan for a warning yielded no results.

r-leyshon: Please ignore above - have followed the trail to the implementation in gtfs_utils.

src/transport_performance/gtfs/multi_validation.py
```diff
@@ -0,0 +1,493 @@
...
+        if isinstance(clean_kwargs, type(None)):
+            clean_kwargs = {}
+        # clean GTFS instances
+        progress = tqdm(
+            zip(self.paths, self.instances), total=len(self.instances)
+        )
+        for path, inst in progress:
+            progress.set_description(f"Cleaning GTFS from path {path}")
+            inst.clean_feed(**clean_kwargs)
+        return None
```
r-leyshon, line 167: Need to test clean_feeds() - we have x2 gtfs fixtures, though they are not in the same location:
tests/data/chester-20230816-small_gtfs.zip
tests/data/gtfs/newport-20230613_gtfs.zip

src/transport_performance/gtfs/multi_validation.py
```diff
@@ -0,0 +1,493 @@
...
+
+        # concat all validation tables into one
+        tables = []
+        for inst in self.instances:
+            valid_df = inst.validity_df.copy()
+            valid_df["GTFS"] = inst.gtfs_path
+            tables.append(valid_df)
+        combined_validation = pd.concat(tables)
+        self.validity_df = combined_validation
+        return self.validity_df.copy().reset_index(drop=True)
```
r-leyshon: Need to test is_valid()

src/transport_performance/gtfs/multi_validation.py
```diff
@@ -0,0 +1,493 @@
...
+        if isinstance(dates, str):
+            dates = [dates]
+        # filter gtfs
+        progress = tqdm(zip(self.paths, self.instances), total=len(self.paths))
+        for path, inst in progress:
+            progress.set_description(f"Filtering GTFS from path {path}")
+            try:
+                inst.filter_to_date(dates=dates)
+            except ValueError as e:
+                raise ValueError(f"In GTFS {path}.\n{e}")
```
r-leyshon, line 274: Testing consider either:
a. Adding an additional small feed fixture with an additional date range that can trigger this condition
b.  Mocking `inst.filter_to_date` in order to return an empty stop_times table, which would trigger the condition.

src/transport_performance/gtfs/multi_validation.py
```diff
@@ -0,0 +1,493 @@
...
+class MultiGtfsInstance:
+    """Create a feed instance for multiple GTFS files.
+
+    This allows for multiple GTFS files to be cleaned, validated, summarised,
+    filtered and saved at the same time.
+
+    Parameters
+    ----------
+    path : Union[str, list]
+        A list of paths, or a glob string. See more informtion on glob strings
```
r-leyshon: glob is awkward to use with pyprojroot.here. Suggest staying with inputs as pathlib.Path / str as with the other functions and using glob within the function to get the list of GTFS files available.

r-leyshon: If it's useful, the snippet I used to get the required gtfs paths out of pyprojroot.here:
`list(here("data/external/gtfs").glob("*.zip"))`

tests/gtfs/test_multi_validation.py
```diff
@@ -0,0 +1,313 @@
...
+    def test_clean_feeds_defences(self, multi_gtfs_fixture):
+        """Defensive tests for .clean_feeds()."""
+        with pytest.raises(TypeError, match=".*clean_kwargs.*dict.*bool"):
+            multi_gtfs_fixture.clean_feeds(True)
+
+    def test_clean_feeds_on_pasas(self):
+        """General tests for .clean_feeds()."""
+        # To be completed once PR 195 is merged as there are breaking changes.
+        # https://github.com/datasciencecampus/transport-network-performance/
+        # pull/195
```
r-leyshon: Due to the size of the other PR it's unlikely to be merged to dev before we need to start running the end to end notebooks. Some testing of the core functionality would be needed.

src/transport_performance/gtfs/multi_validation.py
```diff
@@ -0,0 +1,493 @@
...
+            The kwargs to pass to GtfsInstance.is_valid() for each Gtfs in
+            the MultiGtfsInstance, by default None
+
+        Returns
+        -------
+        self.validity_df : pd.DataFrame
+            A dataframe containing the validation messages from all of the
+            GtfsInstance's.
+
+        """
```
r-leyshon, line 186: When using MultiGtfsInstance.is_valid() for the Marseilles GTFS data available in our GCP bucket, I've needed to use `{"far_stops": False}` in order to avoid 2 separate errors:
1. `ZeroDivisionError: float division by zero` - I believe this is a known issue originating from `validate_travel_over_multiple_stops`
2. `TypeError: 'NoneType' object is not subscriptable` - originating from `validate_travel_between_consecutive_stops`

r-leyshon: The message printed when using `MultiGtfsInstance.is_valid()` states `Cleaning GTFS from path ...` . This challenged my understanding as I just wanted to get the validity table with no cleaning at first, run the cleaning and then re-run `is_valid` to see what changed.

Initially I thought my proposed change would entail updating the message as this isn't a cleaning step, but now it's clear that fast travel cleaners are being used in this step. I'd have thought they would be more appropriate within the `clean_feed` method but have perhaps not understood why this design choice was made.

src/transport_performance/gtfs/multi_validation.py
```diff
@@ -0,0 +1,493 @@
...
+    is_valid()
+        Validates all of the GTFS files.
+    filter_to_date()
+        Filter all of the GTFS files to a specific date(s).
+    filter_to_bbox()
+        Filter all of the GTFS files to a specific bbox.
+    summarise_trips()
+        Create a summary of all of the routes throughout all GTFS files.
+    summarise_routes()
+        Create a summary of all of the trips throughout all GTFS files.
```
r-leyshon, line 65: `viz_stops` not in class docstring

src/transport_performance/gtfs/multi_validation.py
```diff
@@ -0,0 +1,493 @@
...
+        ]
+        progress = tqdm(zip(save_paths, self.instances), total=len(self.paths))
+        for path, inst in progress:
+            progress.set_description(f"Saving at {path}")
+            inst.save(path)
+        return None
+
+    def clean_feeds(self, clean_kwargs: Union[dict, None] = None) -> None:
+        """Clean each of the feeds in the MultiGtfsInstance.
+
```
r-leyshon, line 144: Using this on the GTFS fixtures, I note that X3 trips get removed from route_type 3 on most days.
Would be good to know if the reasons for dropping these.  Are they repeated pair or fast travel drops? Would mean a message stating the route_type (perhaps ID) being dropped due to the following reason. This may not be easy to implement for gtfs_kit cleaning but I understand we are / will be using less of those cleaners.

Before cleaning:
image 2

After cleaning:
image 4

src/transport_performance/gtfs/multi_validation.py
```diff
@@ -0,0 +1,493 @@
...
+    """
+
+    def __init__(self, path: Union[str, list]) -> None:
+        # defences
+        _type_defence(path, "path", (str, list))
+        # defend a glob string
+        if isinstance(path, str):
+            gtfs_paths = glob.glob(path)
+            path = gtfs_paths
```
r-leyshon: For marseilles we have an empty feed. This is because of the folder structure:
```
transport-network-performance % tree data/external/gtfs/UNZIPPED_marseilles_2023-10-24
data/external/gtfs/UNZIPPED_marseilles_2023-10-24
├── ACCM.zip
├── AGGLOBUS.zip
├── AIXENBUS.zip
├── BUSDESCOLLINES.zip
├── CG13.zip
├── CIGALES.zip
├── CIOTABUS.zip
├── COTEBLEUE.zip
├── CPA.zip
├── ETANG.zip
├── FRIOUL.zip
├── LIBEBUS.zip
├── MARCOULINE.zip
├── MILSAB.zip
├── RTM.zip
└── SANPROVENCE.zip

1 directory, 16 files
```

Each of those files are the GTFS archives for different providers. We need an 
init condition that checks the expected GTFS tables are present in the GTFS
directory. If not found, the class needs to unzip the parent folder and access
the GTFS files within.

We should think more about how to do this in a future issue. It could be that
different GTFS sources require a combination of approaches, eg UK style combined
GTFS for one transport mode and the separate service provider files for another
mode.

src/transport_performance/gtfs/multi_validation.py
```diff
@@ -0,0 +1,493 @@
...
+            which="trip", summ_ops=summ_ops
+        ).copy()
+        if return_summary:
+            return self.daily_trip_summary
+
+    def summarise_routes(
+        self,
+        summ_ops: list = [np.min, np.max, np.mean, np.median],
+        return_summary: bool = True,
+    ) -> pd.DataFrame:
```
r-leyshon: Also applies to summarise_trips, row and column sort should be consistent with that of GtfsInstance:
row sort on day is: ['friday', 'friday', 'monday', 'monday', 'saturday', 'saturday', 'sunday', 'sunday', 'thursday', 'thursday', 'tuesday', 'tuesday', 'wednesday', 'wednesday']
column sort is: ['day', 'route_type', 'route_count_max', 'route_count_mean', 'route_count_median', 'route_count_min']
GtfsInstance column sort is ['day', 'route_count_max', 'route_count_mean', 'route_count_median', 'route_count_min', 'route_type']

src/transport_performance/gtfs/multi_validation.py
```diff
@@ -0,0 +1,493 @@
...
+            .reset_index()
+            .round(0)
+        )
+        return total_sums
+
+    def summarise_trips(
+        self,
+        summ_ops: list = [np.min, np.max, np.mean, np.median],
+        return_summary: bool = True,
+    ) -> pd.DataFrame:
```
r-leyshon: Also applies to summarise_routes:
the min / max currently calculated within a GTFS table, then reports max/min from those available feeds.
We need to combine trip / route summary tables then calculate summ stats across all available routes/trips regardless of the GTFS source. Please see Teams chat for further context.

src/transport_performance/gtfs/multi_validation.py
```diff
@@ -0,0 +1,493 @@
...
+            for p in self.paths
+        ]
+        progress = tqdm(zip(save_paths, self.instances), total=len(self.paths))
+        for path, inst in progress:
+            progress.set_description(f"Saving at {path}")
+            inst.save(path)
+        return None
+
+    def clean_feeds(self, clean_kwargs: Union[dict, None] = None) -> None:
+        """Clean each of the feeds in the MultiGtfsInstance.
```
r-leyshon, line 143: Using this method with the Marseilles GTFS available within the GCP bucket, I get lots of the following:
```
Cleaning GTFS from path [.../transport-network-performance/data/external/gtfs/UNZIPPED_marseilles_2023-10-24/RTM.zip](.../transport-network-performance/data/external/gtfs/UNZIPPED_marseilles_2023-10-24/RTM.zip):   6%|▋         | 1/16 [00:00<00:02,  5.56it/s]        
KeyError. Feed was not cleaned.
```
Is this a known error?

src/transport_performance/gtfs/multi_validation.py
```diff
@@ -0,0 +1,493 @@
...
+            The bbox to filter the GTFS to. Leave as none if the GTFS does not
+            need to be cropped. Format - [xmin, ymin, xmax, ymax]
+        crs : str, optional
+            The CRS of the given bbox, by default "epsg:4326"
+
+        Returns
+        -------
+        None
+
+        """
```
r-leyshon: I've tested the new warning on empty feed functionality, and it's working well for single and multiple empty GTFS sources. This is great. I think it would be helpful to raise an error condition if all of the instances are empty, eg you asked for dates or bbox that didn't make any sense for the GTFS you have and now there are nothing but empty feeds.

r-leyshon: Interestingly, calendar_dates for the Marseilles GTFS was always empty for all feeds. Is that expected?

r-leyshon (changes requested):

r-leyshon:
After discussing notebook update with team, we agreed that this PR should create a copy of the e2e marseilles notebook with an example implementation of how to use MultiGtfsInstance.

r-leyshon:
One final request - discussing with the team yesterday, there's a risk that empty feeds may cause problems when running the transport network. A `delete_empty_feed` parameter on the `MultiGtfsInstance` filter methods would safeguard against this. Behaviour would be something like:
* `delete_empty_feed=False` (default), if filtering operation results in any empty feed, raise exception with path to problem GTFS.
*  `delete_empty_feed=True`, if filtering results in any empty feed, warn and delete that feed from the instance.
This should avoid empty feeds being written to file.

src/transport_performance/gtfs/multi_validation.py
cbrowning04: done

src/transport_performance/gtfs/multi_validation.py
cbrowning04: done

src/transport_performance/gtfs/multi_validation.py
cbrowning04: done

src/transport_performance/gtfs/multi_validation.py
cbrowning04: Calendar dates is an optional table

src/transport_performance/gtfs/multi_validation.py
cbrowning04: done

tests/gtfs/test_multi_validation.py
cbrowning04: Done

src/transport_performance/gtfs/multi_validation.py
cbrowning04: done

src/transport_performance/gtfs/multi_validation.py
cbrowning04: Typo in tqdm message. Sorrya about that! Fixed now

cbrowning04:
> One final request - discussing with the team yesterday, there's a risk that empty feeds may cause problems when running the transport network. A `delete_empty_feed` parameter on the `MultiGtfsInstance` filter methods would safeguard against this. Behaviour would be something like:
> 
> * `delete_empty_feed=False` (default), if filtering operation results in any empty feed, raise exception with path to problem GTFS.
> * `delete_empty_feed=True`, if filtering results in any empty feed, warn and delete that feed from the instance.
>   This should avoid empty feeds being written to file.

done

r-leyshon:
@CBROWN-ONS please see patch branch for fix to discovered bug in writing gtfs to file.

One more problem exists before we can merge. The summarise_routes and trips functions are not reporting the expected values for the minimum values:

image 8

image 9

image 10

r-leyshon:
> @CBROWN-ONS please see patch branch for fix to discovered bug in writing gtfs to file.
> 
> One more problem exists before we can merge. The summarise_routes and trips functions are not reporting the expected values for the minimum values:
> 
> image 8
> 
> image 9
> 
> image 10

@CBROWN-ONS thanks for adding in the return_summary functionality. This has allowed me to inspect the above scenario more thoroughly. Can I clarify with you, we are reporting 12 bus routes minimum across all available Mondays in both GTFS feeds. I felt this should be 36 (minimum Newport + minimum Chester). You suggested that it's 12 because there are dates in the summary that only exist for one feed and not the other. I have now confirmed this to be the case.  

```
set(chester_mondays).difference(newport_mondays)
{Timestamp('2024-03-04 00:00:00'),
 Timestamp('2024-03-11 00:00:00'),
 Timestamp('2024-03-18 00:00:00'),
 Timestamp('2024-03-25 00:00:00'),
 Timestamp('2024-04-01 00:00:00'),
 Timestamp('2024-04-08 00:00:00'),
 Timestamp('2024-04-15 00:00:00'),
 Timestamp('2024-04-22 00:00:00')}
```

Available dates in Chester extends to April '24 while Newport stops in March '24. I would suggest the current stat is misleading and the better thing to do would be:

1. Warn that the feeds have differing time ranges and...
2. Filter the dfs to common dates before calculating  summary stats.
3. This would need to work with multiple gtfs sources and considering how to handle differently overlapping date ranges that potentially have no overlap may need some consideration. Keen to get your thoughts on this issue.

r-leyshon:
> > @CBROWN-ONS please see patch branch for fix to discovered bug in writing gtfs to file.
> > One more problem exists before we can merge. The summarise_routes and trips functions are not reporting the expected values for the minimum values:
> > image 8
> > image 9
> > image 10
> 
> @CBROWN-ONS thanks for adding in the return_summary functionality. This has allowed me to inspect the above scenario more thoroughly. Can I clarify with you, we are reporting 12 bus routes minimum across all available Mondays in both GTFS feeds. I felt this should be 36 (minimum Newport + minimum Chester). You suggested that it's 12 because there are dates in the summary that only exist for one feed and not the other. I have now confirmed this to be the case.
> 
> ```
> set(chester_mondays).difference(newport_mondays)
> {Timestamp('2024-03-04 00:00:00'),
>  Timestamp('2024-03-11 00:00:00'),
>  Timestamp('2024-03-18 00:00:00'),
>  Timestamp('2024-03-25 00:00:00'),
>  Timestamp('2024-04-01 00:00:00'),
>  Timestamp('2024-04-08 00:00:00'),
>  Timestamp('2024-04-15 00:00:00'),
>  Timestamp('2024-04-22 00:00:00')}
> ```
> 
> Available dates in Chester extends to April '24 while Newport stops in March '24. I would suggest the current stat is misleading and the better thing to do would be:
> 
> 1. Warn that the feeds have differing time ranges and...
> 2. Filter the dfs to common dates before calculating  summary stats.
> 3. This would need to work with multiple gtfs sources and considering how to handle differently overlapping date ranges that potentially have no overlap may need some consideration. Keen to get your thoughts on this issue.

Input from Ethan:

1. in each instance, calculate number of trips / routes by date and route type.
2. for each date, sum the trips / routes for each route type, across all gtfs. (total of trips/routes by route type per day).
3. Derive summary stats and aggregate.

cbrowning04:
> > > @CBROWN-ONS please see patch branch for fix to discovered bug in writing gtfs to file.
> > > One more problem exists before we can merge. The summarise_routes and trips functions are not reporting the expected values for the minimum values:
> > > image 8
> > > image 9
> > > image 10
> > 
> > 
> > @CBROWN-ONS thanks for adding in the return_summary functionality. This has allowed me to inspect the above scenario more thoroughly. Can I clarify with you, we are reporting 12 bus routes minimum across all available Mondays in both GTFS feeds. I felt this should be 36 (minimum Newport + minimum Chester). You suggested that it's 12 because there are dates in the summary that only exist for one feed and not the other. I have now confirmed this to be the case.
> > ```
> > set(chester_mondays).difference(newport_mondays)
> > {Timestamp('2024-03-04 00:00:00'),
> >  Timestamp('2024-03-11 00:00:00'),
> >  Timestamp('2024-03-18 00:00:00'),
> >  Timestamp('2024-03-25 00:00:00'),
> >  Timestamp('2024-04-01 00:00:00'),
> >  Timestamp('2024-04-08 00:00:00'),
> >  Timestamp('2024-04-15 00:00:00'),
> >  Timestamp('2024-04-22 00:00:00')}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Available dates in Chester extends to April '24 while Newport stops in March '24. I would suggest the current stat is misleading and the better thing to do would be:
> > 
> > 1. Warn that the feeds have differing time ranges and...
> > 2. Filter the dfs to common dates before calculating  summary stats.
> > 3. This would need to work with multiple gtfs sources and considering how to handle differently overlapping date ranges that potentially have no overlap may need some consideration. Keen to get your thoughts on this issue.
> 
> Input from Ethan:
> 
> 1. in each instance, calculate number of trips / routes by date and route type.
> 2. for each date, sum the trips / routes for each route type, across all gtfs. (total of trips/routes by route type per day).
> 3. Derive summary stats and aggregate.

@ethan-moss Are you able to clarify your suggestions here? I'm unclear.

ethan-moss:
> > > > @CBROWN-ONS please see patch branch for fix to discovered bug in writing gtfs to file.
> > > > One more problem exists before we can merge. The summarise_routes and trips functions are not reporting the expected values for the minimum values:
> > > > image 8
> > > > image 9
> > > > image 10
> > > 
> > > 
> > > @CBROWN-ONS thanks for adding in the return_summary functionality. This has allowed me to inspect the above scenario more thoroughly. Can I clarify with you, we are reporting 12 bus routes minimum across all available Mondays in both GTFS feeds. I felt this should be 36 (minimum Newport + minimum Chester). You suggested that it's 12 because there are dates in the summary that only exist for one feed and not the other. I have now confirmed this to be the case.
> > > ```
> > > set(chester_mondays).difference(newport_mondays)
> > > {Timestamp('2024-03-04 00:00:00'),
> > >  Timestamp('2024-03-11 00:00:00'),
> > >  Timestamp('2024-03-18 00:00:00'),
> > >  Timestamp('2024-03-25 00:00:00'),
> > >  Timestamp('2024-04-01 00:00:00'),
> > >  Timestamp('2024-04-08 00:00:00'),
> > >  Timestamp('2024-04-15 00:00:00'),
> > >  Timestamp('2024-04-22 00:00:00')}
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Available dates in Chester extends to April '24 while Newport stops in March '24. I would suggest the current stat is misleading and the better thing to do would be:
> > > 
> > > 1. Warn that the feeds have differing time ranges and...
> > > 2. Filter the dfs to common dates before calculating  summary stats.
> > > 3. This would need to work with multiple gtfs sources and considering how to handle differently overlapping date ranges that potentially have no overlap may need some consideration. Keen to get your thoughts on this issue.
> > 
> > 
> > Input from Ethan:
> > 
> > 1. in each instance, calculate number of trips / routes by date and route type.
> > 2. for each date, sum the trips / routes for each route type, across all gtfs. (total of trips/routes by route type per day).
> > 3. Derive summary stats and aggregate.
> 
> @ethan-moss Are you able to clarify your suggestions here? I'm unclear.

@CBROWN-ONS - sure, no problem! When considering trips/routes for multiple GTFS, the summary descriptive statistics should reflect the case where all the GTFS data where 'one input'. That would then reflect the all services as a whole instead of mins/maxes for each GTFS file - the latter wouldn't give a useful insight into how many services run on days.

To consider all GTFS as 'one input', it can be done in 3 steps:
1. For each GTFS file, calculate the number of trips/routes by date and route_type.
2. For each date, sum trips/routes across all inputs - this is then gives the number of trips/routes by date as if all the GTFS inputs were a single input.
3. Can then aggregate by day of the week.

r-leyshon (approved):

PR #254, opened 2024-02-26
180 cleanup for 195

### LINKED ISSUE #180: fast travel cleaners tech-debt

cbrowning04:
## Description

This issue handles points 3, 4, 5, 6 from #180 

Does not fully fix any issues, only partially #180 
This issue should be marked as closed once #195 is merged.

## Type of change

## How Has This Been Tested?

Test configuration details:
* OS: Windows 10
* Python version: Python 3.9.16
* Java version: N/A
* Python management system: MiniConda

## Checklist:

### DIFF
--- a/src/transport_performance/gtfs/cleaners.py
+++ b/src/transport_performance/gtfs/cleaners.py
@@ -1,5 +1,6 @@
 """A set of functions that clean the gtfs data."""
 from typing import Union
+import warnings
 
 import numpy as np
 
@@ -44,6 +45,12 @@ def drop_trips(gtfs, trip_id: Union[str, list, np.ndarray]) -> None:
         check_elements=True,
         exp_type=str,
     )
+
+    # warn users if passed one of the passed trip_id's is not present in the
+    # GTFS.
+    for _id in trip_id:
+        if _id not in gtfs.feed.trips.trip_id.unique():
+            warnings.warn(UserWarning(f"trip_id '{_id}' not found in GTFS"))
 
     # drop relevant records from tables
     gtfs.feed.trips = gtfs.feed.trips[
--- a/src/transport_performance/gtfs/gtfs_utils.py
+++ b/src/transport_performance/gtfs/gtfs_utils.py
@@ -351,7 +351,6 @@ def filter_gtfs_around_trip(
         An error is raised if a shapeID is not available
 
     """
-    # NOTE: No defence for units as its deleted later on
     _gtfs_defence(gtfs, "gtfs")
     _type_defence(trip_id, "trip_id", str)
     _type_defence(buffer_dist, "buffer_dist", int)
--- a/src/transport_performance/gtfs/gtfs_utils.py
+++ b/src/transport_performance/gtfs/gtfs_utils.py
@@ -5,7 +5,6 @@ from shapely.geometry import box
 from pyprojroot import here
 import pandas as pd
 import os
-import math
 import plotly.graph_objects as go
 from typing import Union, TYPE_CHECKING
 import pathlib
@@ -20,9 +19,10 @@ from transport_performance.utils.defence import (
     _check_iterable,
     _type_defence,
     _check_attribute,
+    _gtfs_defence,
     _validate_datestring,
     _enforce_file_extension,
-    _gtfs_defence,
+    _check_parent_dir_exists,
     _check_item_in_iter,
 )
 from transport_performance.utils.constants import PKG_PATH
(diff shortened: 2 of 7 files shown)


src/transport_performance/gtfs/gtfs_utils.py
```diff
@@ -343,21 +351,22 @@ def filter_gtfs_around_trip(
...
-        if math.isnan(shape_id):
-            raise ValueError(
-                "'shape_id' not available for trip with trip_id: " f"{trip_id}"
-            )
-    except TypeError:
-        pass
+    if pd.isna(shape_id):
+        raise ValueError(
+            "'shape_id' not available for trip with trip_id: " f"{trip_id}"
+        )
```
r-leyshon, line 368: This relates to #180 :
"3. 1 Use pd.isna() in filter_gtfs_around_trip() - handles strings, no need to try/except."

Straight forward  implementation, though I don’t see a test for this. As it's a public function probably for the best. Presume you would need to have a GTFS fixture  with a valid trip ID but no  corresponding shape_id. Take one of our fixtures and lookup a shape_id for a given ID and remove it prior to testing this feature.

tests/utils/test_defence.py
```diff
@@ -251,6 +253,10 @@ def test__gtfs_defence():
         ),
     ):
         _gtfs_defence("tester", "test")
+    # passing test
+    with does_not_raise():
+        gtfs = GtfsInstance(here("tests/data/chester-20230816-small_gtfs.zip"))
+        _gtfs_defence(gtfs=gtfs, param_nm="gtfs")
```
r-leyshon, line 264: This relates to #180 :

"4. Add a unit test for _gtfs_defence() in the passing case (i.e. using a GTFS test fixture)"

* `from contextlib import nullcontext as does_not_raise` is a new one on me. After a bit of googling I found this in the pytest docs. Good to include the link in a comment - https://docs.pytest.org/en/6.2.x/example/parametrize.html#parametrizing-conditional-raising

src/transport_performance/gtfs/cleaners.py
```diff
@@ -45,6 +46,12 @@ def drop_trips(gtfs, trip_id: Union[str, list, np.ndarray]) -> None:
         exp_type=str,
     )
 
+    # warn users if passed one of the passed trip_id's is not present in the
+    # GTFS.
+    for _id in trip_id:
+        if _id not in gtfs.feed.trips.trip_id.unique():
+            warnings.warn(UserWarning(f"trip_id '{_id}' not found in GTFS"))
+
```
r-leyshon, line 54: This relates to #180 :
"6. drop_trip() works with trip_ids that aren't in stop_times. This is OK since no cleaning is done, but it's slightly opaque for public function, maybe we can protect against this? Perhaps raise a warning to say not cleaning since no warnings/no valid trip_ids?"

* Can you tidy up the comment?
* This checks [ID] or IDs against trips.trip_ids. Though the initial ask was to check against stop_times. Check with @ethan-moss as raised the initial issue.
* Should we raise if no valid IDs are passed? If there’s at least one found ID, then warn if user passes IDs not present in trips / stop_times. But if there are no found IDs in trips/stop_times then raise.
* Could use a test in test_cleaners as this is a public function

r-leyshon (changes requested):
Hi @CBROWN-ONS , I've finished up with the review. A couple of queries that I've put in the review comments and something to check with the person raising the initial issue, believe this was @ethan-moss .

~~Also, I couldn't see an implementation for point 5 in #180 :~~

~~"5. For _add_validation_row() add defence to ensure message is only either {"warning", "error"}".~~

~~It may be handled in a different PR, so worth checking.~~

Also please note, I have patched 2 of the CI problems in this branch so do go ahead and check you're happy with that. It was the expired feed fixture thing and then a slight tweak in the requirements file to ensure dask comes with its dataframe dependency.

src/transport_performance/gtfs/gtfs_utils.py
cbrowning04: image
Already implemented. Not covered in the scope of this PR since this is changing previous functionality that does the same thing (test was already written).

tests/utils/test_defence.py
cbrowning04: done

cbrowning04:
Hi @r-leyshon . Point 5 as addressed in you review comment above is fixed in this PR. More specifically, this line of code.
image

tests/utils/test_defence.py
r-leyshon: @CBROWN-ONS have you pushed it up?

src/transport_performance/gtfs/cleaners.py
r-leyshon: I have confirmed  that `gtfs_kit.drop_zombies` is used and handles unmatched trip_ids left hanging in stop_times once `drop_trips` has removed those trip_ids from the trips table.

tests/utils/test_defence.py
cbrowning04: Sorry @r-leyshon . Forgot to push. Should be  up now.

r-leyshon (approved):

PR #294, opened 2024-06-19
287 improve osm maps with tag metadata

### LINKED ISSUE #287: improve OSM maps with tag metadata

Unpacking all tag metadata into custom tooltip for presentation in OSM folium maps

**Additional context**

Discover label metadata for road context, such as speed limits etc

r-leyshon:
## Description

* FindTags now issues a PerformanceWarning if the OSM file it is being instantiated with is larger than 500 KB.
* FindLocations.plot_ids() now has an include_tags parameter, False by default. Setting this to True adds all the tag metadata to the folium map tooltip.
* FindLocations.plot_ids() now exposes selected folium map styling parameters.
* An example implementation is included in the quarto OSM tutorial.
* Finally, all the runners broke when raising PR, as breaking changes with numpy 2.0.0 that had nothing to do with this diff. My solution has been to pin numpy version to that which passed in the previous successful run.

Fixes #287 

## Type of change

## How Has This Been Tested?

Test configuration details:
* OS: macos
* Python version: 3.9.13

## Checklist:


src/transport_performance/osm/validate_osm.py
```diff
@@ -678,11 +700,123 @@ def check_locs_for_ids(self, ids: list, feature_type: str) -> dict:
...
+        tooltip_nm: str = "custom_tooltip",
+        tooltip_kwds: dict = {"labels": False},
+        tiles: str = "CartoDB positron",
+        style_kwds: dict = {
+            "color": "#3f5277",
+            "fill": True,
+            "fillOpacity": 0.3,
+            "fillColor": "#3f5277",
+            "weight": 4,
+        },
```
cbrowning04, line 823: is this the correct approach? it assumes that a user wants to specify all of these features.
In theory, a user might only want to change e.g., the colour, however to do this they would have to specify all the other keywords themselves since style_kwds would be overwritten.

src/transport_performance/osm/validate_osm.py
```diff
@@ -719,6 +873,7 @@ def plot_ids(
         _type_defence(ids, "ids", list)
         _type_defence(feature_type, "feature_type", str)
         _type_defence(crs, "crs", (str, int))
+        _type_defence(include_tags, "include_tags", bool)
```
cbrowning04, line 880: Can you also add defences for the remainder of the new params?

src/transport_performance/osm/validate_osm.py
```diff
@@ -37,6 +39,12 @@
 # ---------utilities-----------
 
 
+class PerformanceWarning(UserWarning):
```
cbrowning04: Do you think inheriting from `Warning` would be better than `UserWarning` here?

src/transport_performance/osm/validate_osm.py
```diff
@@ -541,6 +549,16 @@ def __init__(
         _is_expected_filetype(
             osm_pth, "osm_pth", check_existing=True, exp_ext=".pbf"
         )
+        self.large_file_thresh = 50000  # 50 KB
```
cbrowning04: Very small thing that doesn't really matter.
Do you think this needs to be defined as an attribute of the class? This should be a one time use in the initial size check therefore it can simply be assigned to a local variable.
The only downfall of having it as a class attribute is that a user could potentially view it as something they can alter.

src/transport_performance/osm/validate_osm.py
```diff
@@ -541,6 +549,16 @@ def __init__(
...
+        # implement performance warning on large OSM files.
+        osm_size = os.path.getsize(osm_pth)
+        if osm_size > self.large_file_thresh:
+            warnings.warn(
+                f"PBF file is {osm_size} bytes. Tag operations are expensive."
+                " Consider filtering the pbf file smaller than"
+                f" {self.large_file_thresh} bytes",
+                PerformanceWarning,
+            )
```
cbrowning04: A very small technicality, but the error message provokes the user to reduce file size < 50000, however an error isn't raised if file size is 50000.

src/transport_performance/osm/validate_osm.py
```diff
@@ -678,11 +700,123 @@ def check_locs_for_ids(self, ids: list, feature_type: str) -> dict:
...
+            if not isinstance(d, dict):
+                raise TypeError(f"Expected dict but found {type(d)}: {d}")
+        for id_, tags in dict1.items():  # child_tags is nested
+            # find duplicated keys and prepend parent keys
+            if dupes := set(tags.keys()).intersection(dict2.keys()):
+                for key in dupes:
+                    dict2[f"{prepend_pattern}{key}"] = dict2.pop(key)
+            # merge parent and child tag collections
+            tags_out[id_] = tags | dict2
+        return tags_out
```
cbrowning04, line 741: A very good way of doing this.

cbrowning04 (changes requested):
Great stuff Rich. Just a few small comments on this

docs/tutorials/osm/index.qmd
```diff
@@ -353,6 +353,14 @@ To read more on `osmosis` filtering strategies, refer to the `completeWays` and
 `completeRelations` flag descriptions in the
 [Osmosis detailed usage documentation](https://wiki.openstreetmap.org/wiki/Osmosis/Detailed_Usage_0.48).
 
+
+Note that additional metadata can be added to the map by setting `include_tags=True`. Adding this rich contextual data to the map can be useful but is also expensive. This operation should be avoided for large osm files, for example anything over 500 KB.
```
cbrowning04: 'computationally expensive' rather than 'expensive' here would be a better description, in my opinion, as it provides greater context to the reader.

r-leyshon: Agree

src/transport_performance/osm/validate_osm.py
r-leyshon: It's neat now, but it took a lot of wiggling to get there!

src/transport_performance/osm/validate_osm.py
r-leyshon: It'll still work for sizes greater than 50 KB, it just gets slow on init. It's hoped this would nudge the user to think about the size of the files prior to using FindTags as it's the slowest of all the osm flavoured classes.

src/transport_performance/osm/validate_osm.py
r-leyshon: agree with your point. I used name mangling and updated the docstring.

src/transport_performance/osm/validate_osm.py
r-leyshon: agree

src/transport_performance/osm/validate_osm.py
r-leyshon: agree. good shout.

src/transport_performance/osm/validate_osm.py
r-leyshon: I think this one is ok:
image

r-leyshon:
Hey @CBROWN-ONS , I've just implemented your suggestions and pushed the diff to the PR. Thanks for your review on this, some good catches in there.

cbrowning04 (approved):

cbrowning04:
Thanks for the changes @r-leyshon . All good, will merge this now.

